PR #3, opened 2019-03-29
Update 1

tokyoraven02:
Added UEFI support, AMD and Intel microcodes, NTFS driver, tweaked packages to install, changed mount point to /mnt, GPU detection and appropriate driver installation.


install.sh
```diff
@@ -5,7 +5,7 @@ set -e
 export USERNAME=gamer
 export SYSTEM_NAME=gamer-os
 
-export MOUNT_PATH=/tmp/${SYSTEM_NAME}
+export MOUNT_PATH=/mnt
```
alkazar: someone might already have things mounted in /mnt

install.sh
```diff
@@ -29,25 +29,50 @@ if [[ ! $REPLY =~ ^[Yy]$ ]]; then
...
+	grub-mkconfig -o /boot/grub/grub.cfg
+else
+	pacman --noconfirm -S \
+	grub
+	grub-install --target=i386-pc ${DISK}
+	grub-mkconfig -o /boot/grub/grub.cfg
+fi
 
 # enable bluetooth connection for xbox one s controller
 sed -i 's/^GRUB_CMDLINE_LINUX=.*/GRUB_CMDLINE_LINUX="bluetooth.disable_ertm=1"/' /etc/default/grub
```
alkazar: this line needs to execute before grub-mkconfig

install.sh
```diff
@@ -56,8 +81,10 @@ Include = /etc/pacman.d/mirrorlist
 pacman --noconfirm -Sy
 pacman --noconfirm -S \
 	lightdm \
+	lightdm-gtk-greeter \
```
alkazar: why install this?

install.sh
```diff
@@ -56,8 +81,10 @@ Include = /etc/pacman.d/mirrorlist
...
 	accountsservice \
 	xorg-server \
+	xorg-apps \
```
alkazar: what is this needed for?

install.sh
```diff
@@ -71,15 +98,24 @@ pacman --noconfirm -S \
 	python \
 	vulkan-icd-loader \
 	lib32-vulkan-icd-loader \
-	vulkan-radeon \
-	lib32-vulkan-radeon \
-	steam
+	steam \
+	steam-native-runtime
```
alkazar: since we are not using this by default, no point in installing this. if someone wants to use it, they can install manually and modify the steamos session. i am open to using the native runtime by default, but when i tested this the majority of my games wouldnt even start.

install.sh
```diff
@@ -71,15 +98,24 @@ pacman --noconfirm -S \
...
 if lspci | grep -E -i '(vga|3d)' | grep -i nvidia > /dev/null; then
 	pacman --noconfirm -S nvidia nvidia-utils lib32-nvidia-utils
 fi
 
+# install AMD graphics
+if lspci | grep -E -i '(vga|3d)' | grep -i AMD > /dev/null; then
+	pacman --noconfirm -S vulkan-radeon lib32-vulkan-radeon xf86-video-amdgpu lib32-mesa libva-mesa-driver lib32-libva-mesa-driver mesa-vdpau lib32-mesa-vdpau
+fi
+
+# install Intel graphics
```
alkazar: i am not sure how safe this is to be installed alongside nvidia drivers. the arch wiki says it can cause problems.

alkazar:
Could you please submit a single PR for each logical change? that makes it easier to test, review and merge independently.

For example, one PR for UEFI support, one for graphics driver changes, etc.

Another example is I might like the UEFI changes, but disagree with some of the other package additions you made. This prevents me from merging the parts I like.

install.sh
tokyoraven02: No as it states it checks which GPU is using and then install drivers for appropriate GPU

install.sh
tokyoraven02: Its a part of X.org server that contains utilities which could potentially fix some game issues, due to lack of utilities that games require.

install.sh
tokyoraven02: I removed it from project after some research, maybe in future it will be much better to use !

install.sh
tokyoraven02: Which line does it say ? I can't see a problem here.

PR #5, opened 2019-03-31
Added automatic timezone detection

tokyoraven02:
Added automatic timezone detection through IP address

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -62,6 +62,37 @@ Include = /etc/pacman.d/mirrorlist
 " >> /etc/pacman.conf
 
 pacman --noconfirm -Sy
+
+if lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: amdgpu' > /dev/null; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -51,7 +51,7 @@ else
 	mount ${DISK}1 ${MOUNT_PATH}
 fi
 
-pacstrap ${MOUNT_PATH} base
+pacstrap ${MOUNT_PATH} base ntp
```
alkazar: why here and not just install it as part of the other packages we install?

install.sh
```diff
@@ -116,6 +116,12 @@ autologin-user=${USERNAME}
 autologin-session=steamos
 " > /etc/lightdm/lightdm.conf
 
+TIMEZONE="$(curl -s https://ipapi.co/timezone)"
```
alkazar: what happens when that curl call fails? can we handle that case better?

PR #6, opened 2019-03-31
added nvidia's opencl package

alwayslivid:
This package is not really that useful, but still useful up to a point nevertheless.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -87,7 +87,7 @@ pacman --noconfirm -S \
 
 # install nvidia graphics if needed
 if lspci | grep -E -i '(vga|3d)' | grep -i nvidia > /dev/null; then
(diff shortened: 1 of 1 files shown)


alkazar:
@alwayslivid is this useful for gaming at all?

alwayslivid:
@alkazar Although OpenCL is typically used in applications such as Adobe's Photoshop and GIMP, there are some legacy games that utilize OpenCL, mostly as a physics engine.

A notable example to this has to be (BeamNG.drive)[https://www.beamng.com/], even though there's no native Linux support for it.

alkazar:
included in 992687b3c6a520dc0b2d4fc36081aa9f79e53a1e

PR #7, opened 2019-03-31
GPU detection/driver installation.

alwayslivid:
The hard work was done by @tokyoraven02, but I decided to change up the code a bit so that it'd align with the indentation of the mainstream code.

I also worked on the readability of the code and the messages that appear on the screen.

I believe that this is ready to go.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -68,6 +68,7 @@ pacman --noconfirm -Sy
 
 # basic package installation
 pacman --noconfirm -S \
+	linux-headers \
 	lightdm \
 	accountsservice \
 	xorg-server \
@@ -106,22 +107,32 @@ if echo "$devices" | grep -i 'NVIDIA Corporation' > /dev/null; then
 
 # AMD
 elif echo "$devices" | grep -i 'AMD' > /dev/null; then
-	echo "AMD GPU with radeon kernel detected, switching kernel module and installing drivers..."
-	echo "blacklist radeon" > /etc/modprobe.d/blacklist.conf
-	echo "options amdgpu si_support=1" > /etc/modprobe.d/amdgpu.conf
-	echo "options amdgpu cik_support=1" >> /etc/modprobe.d/amdgpu.conf
-	echo "options radeon si_support=0" > /etc/modprobe.d/radeon.conf
-	echo "options radeon cik_support0" >> /etc/modprobe.d/radeon.conf
-	pacman --noconfirm -S \
-		vulkan-radeon mesa \
-		vulkan-icd-loader \
-		xf86-video-amdgpu \
-		libva-mesa-driver \
-		mesa-vdpau \
-		lib32-libva-mesa-driver \
-		lib32-mesa-vdpau \
-		lib32-mesa \
-		lib32-vulkan-radeon
+	if lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: radeon, amdgpu'; then
+		echo "AMD GPU with radeon kernel detected, switching kernel module and installing drivers..."
+		echo "blacklist radeon" > /etc/modprobe.d/blacklist.conf
+		echo "options amdgpu si_support=1" > /etc/modprobe.d/amdgpu.conf
+		echo "options amdgpu cik_support=1" >> /etc/modprobe.d/amdgpu.conf
+		echo "options radeon si_support=0" > /etc/modprobe.d/radeon.conf
+		echo "options radeon cik_support0" >> /etc/modprobe.d/radeon.conf
+		pacman --noconfirm -S \
+			vulkan-radeon mesa \
+			vulkan-icd-loader \
+			xf86-video-amdgpu \
+			libva-mesa-driver \
+			mesa-vdpau \
+			lib32-libva-mesa-driver \
+			lib32-mesa-vdpau \
+			lib32-mesa \
+			lib32-vulkan-radeon
+	else
+		echo "Legacy AMD GPU detected, installing drivers..."
+		pacman --noconfirm -S \
+			mesa \
+			lib32-mesa \
+			xf86-video-ati \
+			mesa-vdpau \
+			lib32-mesa-vdpau
+	fi
 
 # Intel
 elif echo "$devices" | grep -i 'Intel Corporation' > /dev/null; then
--- a/install.sh
+++ b/install.sh
@@ -62,6 +62,37 @@ Include = /etc/pacman.d/mirrorlist
 " >> /etc/pacman.conf
 
 pacman --noconfirm -Sy
+
+if lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: amdgpu' > /dev/null; then
+  echo "You are using AMD Graphics card with amdgpu kernel module, we are going install open source drivers"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -85,13 +90,59 @@ pacman --noconfirm -S \
...
+elif lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: radeon' > /dev/null; then
+	echo "Legacy AMD GPU detected, installing drivers..."
+	pacman --noconfirm -S \
+		mesa \
+		lib32-mesa \
+		xf86-video-ati \
+		mesa-vdpau \
+		lib32-mesa-vdpau
+
+elif lspci | grep -E -i '(vga|3d)' | grep -i 'Intel Corporation' > /dev/null; then
```
alkazar: should intel go at the end? I am thinking of the case of dual graphics.

install.sh
```diff
@@ -51,17 +51,22 @@ else
 	mount ${DISK}1 ${MOUNT_PATH}
 fi
 
+# chroot into target
 pacstrap ${MOUNT_PATH} base
-arch-chroot ${MOUNT_PATH} /bin/bash <<EOF
```
alkazar: will the commands below this line be executed within the chroot without the redirect? if it does work, we probably need to remove the `EOF` at the very end of this file.

install.sh
```diff
@@ -51,17 +51,22 @@ else
...
 echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen
 locale-gen
 
+# adding multilib to pacman mirror list
 echo "
 [multilib]
 Include = /etc/pacman.d/mirrorlist
 " >> /etc/pacman.conf
 
+# system update
```
alkazar: well, not really an update, we need to do this to grab the multilib package database; perhaps the comment could say `update package datbases` ?

install.sh
```diff
@@ -85,13 +90,59 @@ pacman --noconfirm -S \
...
+	echo "options radeon cik_support0" >> /etc/modprobe.d/radeon.conf
+	pacman --noconfirm -S \
+		vulkan-radeon mesa \
+		vulkan-icd-loader \
+		xf86-video-amdgpu \
+		libva-mesa-driver \
+		mesa-vdpau \
+		lib32-libva-mesa-driver \
+		lib32-mesa-vdpau \
+		lib32-mesa \
```
alkazar: isn't this already installed with steam?

install.sh
```diff
@@ -85,13 +90,59 @@ pacman --noconfirm -S \
...
+		lib32-vulkan-radeon
+
+elif lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: radeon' > /dev/null; then
+	echo "Legacy AMD GPU detected, installing drivers..."
+	pacman --noconfirm -S \
+		mesa \
```
alkazar: isn't `mesa` and `lib32-mesa` already installed with steam?

install.sh
```diff
@@ -85,13 +90,59 @@ pacman --noconfirm -S \
...
 
-# install nvidia graphics if needed
-if lspci | grep -E -i '(vga|3d)' | grep -i nvidia > /dev/null; then
-	pacman --noconfirm -S nvidia nvidia-utils lib32-nvidia-utils
-fi
-
 systemctl enable NetworkManager lightdm bluetooth
 
+# gpu detection/driver installation
+elif lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: radeon, amdgpu' > /dev/null; then
```
alkazar: will this ever be true when installing from the Arch linux installer which is command line only?

install.sh
```diff
@@ -85,13 +90,59 @@ pacman --noconfirm -S \
...
+		vulkan-icd-loader \
+		xf86-video-amdgpu \
+		libva-mesa-driver \
+		mesa-vdpau \
+		lib32-libva-mesa-driver \
+		lib32-mesa-vdpau \
+		lib32-mesa \
+		lib32-vulkan-radeon
+
+elif lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: radeon' > /dev/null; then
```
alkazar: Same comment: will this ever be true when installing from the Arch linux installer which is command line only?

install.sh
```diff
@@ -116,15 +168,21 @@ autologin-user=${USERNAME}
 autologin-session=steamos
 " > /etc/lightdm/lightdm.conf
 
+TIMEZONE="$(curl -s https://ipapi.co/timezone)"
```
alkazar: this is part of another PR. Please remove. We haven't quite decided what to do for timezone yet.

install.sh
```diff
@@ -85,13 +90,59 @@ pacman --noconfirm -S \
...
-if lspci | grep -E -i '(vga|3d)' | grep -i nvidia > /dev/null; then
-	pacman --noconfirm -S nvidia nvidia-utils lib32-nvidia-utils
-fi
-
 systemctl enable NetworkManager lightdm bluetooth
 
+# gpu detection/driver installation
+elif lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: radeon, amdgpu' > /dev/null; then
+	echo "AMD GPU with radeon kernel detected, switching kernel module and installing drivers..."
+	echo "blacklist radeon" > /etc/modprobe.d/blacklist.conf
```
alkazar: perhaps combine `blacklist.conf` and `amdgpu.conf` into one `gameros.conf`?

install.sh
alwayslivid: The Intel part should not be executed, should it stay in the end, since this is an if else statement.

install.sh
alwayslivid: @tokyoraven02

install.sh
alwayslivid: The script installs the GPU drivers first.

install.sh
alwayslivid: If we're going to do this, it'd be a good idea to pull this off at a later stage.

install.sh
alwayslivid: The driver installation process comes before the Steam compositor installation.

install.sh
alwayslivid: This issue has been temporarily dealt with.

We need to test this out.

install.sh
alwayslivid: @tokyoraven02 wrote this feature with the Arch Linux wiki in mind.


This is a reasonable concern, albeit I don't believe that the fact that this script is supposed to be executed within the Arch Linux installer is going to make a difference, since we're talking about kernel modules over here.

We are in need of testers.

install.sh
alwayslivid: We have to test this out. The thing I said earlier about kernel modules still applies here.

alwayslivid:
There are a few things that need testing.

Namely,
- GPU detection for AMD/Radeon cards. @tokyoraven02 
- does the removal of the EOF parts break the script?

install.sh
alkazar: why?

alwayslivid:
Me and @tokyoraven02 worked on the GPU detection script.

alkazar:
I ended up having to modify this PR quite a bit to get it working. I have committed myself including @alwayslivid and @tokyoraven02 as co-authors.

PR #9, opened 2019-04-03
Fixed GPU support

tokyoraven02:
Quick fix to some mistakes [as missing mesa driver, Added option to detect modern AMD Graphics cards like Vega, Polaris, Raven Ridge (Ryzen APUs)]

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -112,8 +112,6 @@ elif lspci | grep -E -i '(vga|3d|display)' | grep -i amd > /dev/null; then
 		echo "options radeon si_support=0" >> /etc/modprobe.d/gameros.conf
 		echo "options radeon cik_support=0" >> /etc/modprobe.d/gameros.conf
 		pacman --noconfirm -S \
-			mesa \
-			lib32-mesa \
 			vulkan-icd-loader \
 			lib32-vulkan-icd-loader \
 			libva-mesa-driver \
@@ -127,8 +125,6 @@ elif lspci | grep -E -i '(vga|3d|display)' | grep -i amd > /dev/null; then
 	elif lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: amdgpu' > /dev/null; then
 		echo "AMD graphics card with amdgpu kernel module detected, installing drivers..."
 		pacman --noconfirm -S \
-			mesa \
-			lib32-mesa \
 			vulkan-icd-loader \
 			lib32-vulkan-icd-loader \
 			libva-mesa-driver \
@@ -142,8 +138,6 @@ elif lspci | grep -E -i '(vga|3d|display)' | grep -i amd > /dev/null; then
 	else
 		echo "Legacy AMD/ATI GPU detected, installing drivers..."
 		pacman --noconfirm -S \
-			mesa \
-			lib32-mesa
 			xf86-video-ati \
 			mesa-vdpau \
 			lib32-mesa-vdpau
--- a/install.sh
+++ b/install.sh
@@ -105,13 +105,15 @@ if lspci | grep -E -i '(vga|3d|display)' | grep -i nvidia > /dev/null; then
 # AMD
 elif lspci | grep -E -i '(vga|3d|display)' | grep -i amd > /dev/null; then
 	if lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: radeon, amdgpu' > /dev/null; then
-		echo "AMD GPU with radeon kernel detected, switching kernel module and installing drivers..."
+		echo "AMD Graphics card with radeon kernel module detected, switching kernel module and installing drivers..."
 		echo "blacklist radeon" >> /etc/modprobe.d/gameros.conf
 		echo "options amdgpu si_support=1" >> /etc/modprobe.d/gameros.conf
 		echo "options amdgpu cik_support=1" >> /etc/modprobe.d/gameros.conf
 		echo "options radeon si_support=0" >> /etc/modprobe.d/gameros.conf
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -105,13 +105,15 @@ if lspci | grep -E -i '(vga|3d|display)' | grep -i nvidia > /dev/null; then
...
-		echo "AMD GPU with radeon kernel detected, switching kernel module and installing drivers..."
+		echo "AMD graphics card with radeon kernel module detected, switching kernel module and installing drivers..."
 		echo "blacklist radeon" >> /etc/modprobe.d/gameros.conf
 		echo "options amdgpu si_support=1" >> /etc/modprobe.d/gameros.conf
 		echo "options amdgpu cik_support=1" >> /etc/modprobe.d/gameros.conf
 		echo "options radeon si_support=0" >> /etc/modprobe.d/gameros.conf
 		echo "options radeon cik_support=0" >> /etc/modprobe.d/gameros.conf
 		pacman --noconfirm -S \
+			mesa \
+			lib32-mesa \
```
alkazar: this is unnecessary, these packages are already installed with the base set of packages

install.sh
```diff
@@ -105,13 +105,15 @@ if lspci | grep -E -i '(vga|3d|display)' | grep -i nvidia > /dev/null; then
 # AMD
 elif lspci | grep -E -i '(vga|3d|display)' | grep -i amd > /dev/null; then
 	if lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: radeon, amdgpu' > /dev/null; then
```
alkazar: not sure why `amdgpu` is listed here as well, is this right?

install.sh
```diff
@@ -121,9 +123,27 @@ elif lspci | grep -E -i '(vga|3d|display)' | grep -i amd > /dev/null; then
 			vulkan-radeon \
 			lib32-vulkan-radeon \
 			xf86-video-amdgpu
+	
+	elif lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: amdgpu' > /dev/null; then
+		echo "AMD graphics card with amdgpu kernel module detected, installing drivers..."
+		pacman --noconfirm -S \
```
alkazar: would be nice to combine this with the list above so we don't have to repeat ourselves. unless we expect this list to diverge in the future?

install.sh
```diff
@@ -121,9 +123,27 @@ elif lspci | grep -E -i '(vga|3d|display)' | grep -i amd > /dev/null; then
...
+			lib32-mesa-vdpau \
+			vulkan-radeon \
+			lib32-vulkan-radeon \
+			xf86-video-amdgpu		
+
 	else
 		echo "Legacy AMD/ATI GPU detected, installing drivers..."
 		pacman --noconfirm -S \
+			mesa \
+			lib32-mesa
```
alkazar: not needed

install.sh
```diff
@@ -121,9 +123,27 @@ elif lspci | grep -E -i '(vga|3d|display)' | grep -i amd > /dev/null; then
 			vulkan-radeon \
 			lib32-vulkan-radeon \
 			xf86-video-amdgpu
+	
+	elif lspci -nnk | grep -i vga -A3 | grep 'Kernel modules: amdgpu' > /dev/null; then
+		echo "AMD graphics card with amdgpu kernel module detected, installing drivers..."
+		pacman --noconfirm -S \
+			mesa \
+			lib32-mesa \
```
alkazar: mesa and lib32-mesa are not needed here

install.sh
tokyoraven02: Yes, it's a new kernel module that replaced radeon kernel module some time ago.

install.sh
tokyoraven02: No they it isnt, if it is then it will include 64-bit mesa and ignore 32 bit mesa one which is a must have for proton

install.sh
tokyoraven02: First one is Sea Island and Southern Island, we are not repeating cause modern AMD Graphics cards use only amdgpu kernel module and don't work with radeon module. There are big changes in way that they are built in Linux kernel.

install.sh
tokyoraven02: For Proton are required, I will re tweak base packages

install.sh
tokyoraven02: Again, for proton games

install.sh
alkazar: I ran the installer without,the gpu section and both 64 and 32 bit mesa libraries were installed

install.sh
alkazar: this condition and thecone above contain installation of the exact same packages. you could move the package installation up into a condition that catches both gpu cases and then do the configuration only in the one case. that way we dont have to repeat the  list of packages to be installed.

install.sh
alkazar: yes, but I mean, why is amdgpu string included when detecting radeon kernel module. Seems confusing. I thought we were detecting for amdgpu here until you added the new condition.

tokyoraven02: There are 3 types of AMD Graphics cards of Linux :
1) Legacy / ATI (They use radeon kernel module)
2) Sea Island and Southern Island (They use radeon kernel module by default but its recommended to switch to amdgpu for better performance in Proton / Native games)
3) Modern GPUs - Polaris, Vega, Raven Ridge (They use amdgpu kernel module by default)

This detection is detecting all 3 types of those cards, we don't want to confuse users with wrong information

install.sh
tokyoraven02: It cannot be matched since modern AMD graphics cards don't use radeon kernel module, it would result in error and skipping driver installation.

PR #19, opened 2019-07-21
Added scripts for building ISOs

sharkwouter:
This adds a script to build an ISO from either Arch or a Docker container. It uses archiso for building ISO files. Instructions on how to use archiso can be found here: https://wiki.archlinux.org/index.php/Archiso

The gamer-os directory is a mostly unchanged copy of the releng archiso profile right now. The only change is that I removed the install.txt file from the home of the root user. The build script adds the install.sh file to it when run.

More changes should still be made in the future.


gamer-os/packages.x86_64
```diff
@@ -0,0 +1,66 @@
+arch-install-scripts
```
alkazar: is there a way to also include the frzr AUR package? Either built from the AUR (preferred) or pre-built package which is currently available here: https://github.com/gamer-os/frzr/releases

docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
+#Dockerfile for building VaporOS Desktop
```
alkazar: can we update the comment to reference `gamerOS` instead of `VaporOS`?

gamer-os/packages.x86_64
sharkwouter: This can be done. Archiso uses the airootfs directory as base for its root filesystem of the live environment which you boot to when booting from the ISO it generates. The script root/customize_airootfs.sh is run before finalizing the ISO.

What you can do is adding the package to the airootfs directory and then change the customize_airootfs.sh script to install it using pacman. That way frzr will be available on the live environment booted from the ISO.

Another option would be to host frzr in a package repository which can be used by pacman. Then you can add your repo to the pacman.conf file and the package name to the packages.x86_64 file.

sharkwouter:
> Could we put all the installer related files under a single directory, like `install-media` but preferably something better (lol)? or what do you think of making a separate repo for this (that would also include the install.sh) script. The gamer-os repo would become only a repo for building the gamerOS images themselves. I think this would make it easier for people to fork and create their on gamerOS versions, but we can share the installer iso since it would be exactly the same (except for specifying a channel which we could allow users to do in the installer as an option).

I think it would probably be easier to maintain if the system for building the ISO and the installer itself would be kept in the same repo. In that I am making the assumption that not many people would just want to download the script and people download the ISO instead.

Right now the build script is copying the install.sh file to the correct place in the airootfs directory before building. I think it would be better to move it there in the git repo instead, but that would make the instructions currently in the README stop working.

alkazar:
this PR was moved to the `install-media` repo. Closing.

PR #31, opened 2019-08-20
Add trim support

sharkwouter:
This unit for periodical trim is installed with the base group, so it only has to be activated. You can read up on it here: https://wiki.archlinux.org/index.php/Solid_state_drive#Periodic_TRIM

### DIFF
--- a/profiles/default
+++ b/profiles/default
@@ -77,5 +77,5 @@ export SERVICES="\
 	NetworkManager \
 	lightdm \
 	bluetooth \
-	fs-trim.timer \
+	fstrim.timer \
 "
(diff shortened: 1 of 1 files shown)


alkazar:
I thought there also needs to be some special mount options added to fstab for trim to work?

sharkwouter:
Well, for trim support you can go one of 2 ways. You can periodically trim or trim continuously. This method uses trim once per week and it only does so on disks which support it.

Trimming continuously can cause performance issues and in some cases can cause the SSD to degrade faster. If you want to use continuously trim, you'll also need to make sure you know which drive supports it before mounting. In other words, this solution is a bit safer and easier.

PR #35, opened 2019-08-23
UEFI support

sharkwouter:
This makes gamer-os able to mount the new boot partition and update the system later.


sharkwouter:
The changes to profiles/default will actually not be needed if the dependencies for frzr are updated to include efibootmgr and dosfstools in the future.

sharkwouter:
I've merged it with master to prevent the merge conflict.

sharkwouter:
I'm going to close this one in favor of the Plymouth pull request, since it also contains these commits.

PR #36, opened 2019-08-23
Update lsb_release output based on the channel and version used for build.sh

sharkwouter:
This is probably the easiest way to get this working. Now the channel and the version specified when running build.sh are used for the release which Steam displays. Replacing DISTRIB_DESCRIPTION with a better looking version of the name might be nice as well, but I'll leave that up to you.


alkazar:
Sorry, I already had this functionality in, I just forgot to push it.

PR #37, opened 2019-08-24
UEFI and Plymouth

sharkwouter:
This adds plymouth and updates the mkinitcpio to use it. It also includes the UEFI patch.


sharkwouter:
I've merged it with master to prevent the merge conflict.

sharkwouter:
This should now work with BIOS systems without breaking backwards comparability. I'm still testing this at the moment.

sharkwouter:
I removed efibootmgr and dosfstools from the package list, since frzr will have these as dependencies. This does is likely to break UEFI systems until frzr is actually updated in the AUR to have these dependencies.

PR #40, opened 2019-08-29
Removed some packages which aren't used

sharkwouter:
Using htop and flatpak is currently not doable from within Steam. Python should be a dependency.

### DIFF
--- a/profiles/default
+++ b/profiles/default
@@ -3,7 +3,6 @@
 export SIZE="10GB"
 
 export PACKAGES="\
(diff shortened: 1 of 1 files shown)


alkazar:
- I use htop for recovering when games hang or other things go wrong
- python is needed for proton but is not a dependency on steam
- flatpak I want to leave so power users can install additional software

sharkwouter:
Makes sense, thanks for the detailed explanation.

PR #57, opened 2019-12-08
Support changing lights on Alienware Alpha

sharkwouter:
This makes you able to select Alien FX in the menu if you have an Alienware Alpha or Steam Machine to change the color of the lights on the system.

I've created a package on the AUR which takes the script from the steamos-base-files. This is what is used here.

I did find a bug while working on this. The files in ``/etc/sudoers.d`` aren't loaded by sudo because ``build.sh`` adds a line below the include line in the sudoers file. With this pull request this is fixed by writing to the second to last line instead.


sharkwouter:
AUR link of the package: https://aur.archlinux.org/packages/alienware-alpha-wmi

PR #61, opened 2020-01-06
Added Xbox One Wireless adapter support

tommy4st:
This change adds support for the Xbox One Wireless adapter by [medusalix](https://github.com/medusalix/xow).

It's tested on my box. The only problem is, after a restart you have to pair your controllers again. I'm not sure how to fix this, but the the xow project is very active at the moment, so maybe it's an upstream fix.

P.S. For me the re-pair problem only occurs, when the controllers are already connected during a reboot.

### DIFF
--- a/profiles/default
+++ b/profiles/default
@@ -88,6 +88,7 @@ export AUR_PACKAGES="\
 	steam-buddy \
 	retroarch-autoconfig-udev-git \
 	alienware-alpha-wmi \
+	xow-git \
 "
 
 export SERVICES="\
@@ -102,4 +103,5 @@ export SERVICES="\
 	steam-buddy-proxy@gamer.service \
 	steam-buddy-proxy@gamer.socket \
 	haveged \
+	xow \
 "


alkazar:
Oh, nice. Thanks for the PR.

I have been following this project and hoped to add it at some point.

I do have a few concerns before I do though.

1. does adding this preclude connecting via bluetooth?
2. how does pairing work? Through the Big picture bluetooth UI?
3. how is the button mapping? does it work well in games? I wouldn't expect games to properly detect it yet. In particular, does it work with the retroarch emulators that come pre-installed in GamerOS?

Last I checked the project wasn't quite stable yet, it even comes with a warning. I was planning on waiting a bit more before adding it, but I could be convinced to add it earlier.

Disconnecting bluetooth controllers is a known issue: https://github.com/ValveSoftware/steam-for-linux/issues/6756

I wouldn't have expected it to be an issue with the dongle and am unsure if it is actually related to the issue listed above.

tommy4st:
> 1. does adding this preclude connecting via bluetooth?
Using the wireless dongle doesn't depend on Bluetooth at all, it's more like a weird configured Wifi dongle. So you only connect your controllers to the dongle. Other Bluetooth devices are still usable.

> 2. how does pairing work? Through the Big picture bluetooth UI?
Pairing works like with the 360 wireless dongle. You push a button on the dongle and the controller to initiate pairing. It works exactly like in Windows.

> 3. how is the button mapping? does it work well in games? I wouldn't expect games to properly detect it yet. In particular, does it work with the retroarch emulators that come pre-installed in GamerOS?
In Steam the mapping is as expected. I still need to test it with retroarch. (maybe I manage to do this in the evening)

> Last I checked the project wasn't quite stable yet, it even comes with a warning. I was planning on waiting a bit more before adding it, but I could be convinced to add it earlier.
I tested it on my Arch machine with two controllers and it worked perfectly, even the rumbling. That's why I tested it with GamerOS with the same result (two controllers connected, rumble working). As it is not directly coupled to the kernel it should not pose a threat to overall system stability. Probably the worst thing which can happen is that the controllers don't work. As it is a quite small program, it shouldn't occupy much of the system resources.

> Disconnecting bluetooth controllers is a known issue: [ValveSoftware/steam-for-linux#6756](https://github.com/ValveSoftware/steam-for-linux/issues/6756)
> I wouldn't have expected it to be an issue with the dongle and am unsure if it is actually related to the issue listed above.
It's a different issue. I did some more tests and I need to file an issue with the project. It only occurs, when I restart and the controllers are still switched on. When I activate a controller after starting up, it works like a charm, but not if it was already activated before. For me the benefits still outweigh the disadvantages by a huge margin, because the connection is now much more stable than with bluetooth.

alkazar:
Thank you for your detailed reply. Sounds good.

PR #65, opened 2020-01-12
added mount of etc to initrd

### LINKED ISSUE #60: /etc overlay established too late?
Some changes to /etc are not effective. Probably because the overlay is mounted to late in the process.

For example I cannot add a secondary hard drive to fstab. I can actually add it, but it is not mounted because fstab is read before the overlay is up. The problem is not about the directory, because I just created on in the home directory and I can mount it later on manually, but not on startup.

Another example is systemd services: I cannot enable a service. Again I can actually enable it in the way that the link is created, but it is not processed.

GamerOS 10, default

tommy4st:
Fixes #60 (maybe)

Not tested yet! (but it builds without new errors)

The following error is there, but it's not related to changes by this PR:
```
==> ERROR: module not found: `aes_x86_64'
```


tommy4st:
Now it's tested and after fixing some stupid errors it's working for me :grin:

tommy4st:
Sorry for all the changes, but I think this way the build process is a bit more straightforward. The rootfs directory now holds all static files, which were previously generated by the build.sh script. This way escaping of some stuff is not an issue anymore (at least for me).

The mkinitcpio program runs multiple times, when I build the image. This is still the case and I don't know why. Also I moved the configuration to it's default path at /etc. The config is also not deleted anymore, thereby you can recreate initrd when unlocking your installation.

This build is tested and working as expected.

alkazar:
Great work, thanks a lot!

rootfs/etc/mkinitcpio.conf
```diff
@@ -0,0 +1,6 @@
+# vim:set ft=sh
+
+MODULES=(i915 dm_mod ext4 sha256 sha512 overlay)
```
alkazar: why add the `i915` module here?

tommy4st: Sorry, it wasn't supposed to be in here. I wanted to do the "flicker free boot experience" like it is documented here: [Arch Wiki: Plymouth](https://wiki.archlinux.org/index.php/Plymouth#The_plymouth_hook) And as I understand it, these modules are only optional and the fallback image - which is created by default, but you are not using - contains all available modules. So i figured it won't hurt. But you are obviously free to remove it :grin:

PR #81, opened 2020-03-24
Add policykit action and rule to allow SteamVR to setcap on vrcompositor

ShadowApex:
When launching SteamVR, you will be prompted for sudo rights to run `setcaps` on `vrcompositor-launcher`, but GamerOS does not have a policykit agent running to display a password prompt:

```
    if ! ${STEAM_ZENITY} --no-wrap --question --text="SteamVR requires superuser access to finish setup. Proceed?"; then
        pErr 'Error: user declined superuser request.'
        return 1
    fi

    pkexec setcap CAP_SYS_NICE=eip $STEAMVR_TOOLSDIR/bin/linux64/vrcompositor-launcher
```

This PR adds a policykit action and rule to allow sudoless access for `vrstartup.sh` to set this capability.

### DIFF
--- a/rootfs/etc/polkit-1/rules.d/41-steamvr.rules
+++ b/rootfs/etc/polkit-1/rules.d/41-steamvr.rules
@@ -1,0 +1,10 @@
+/* Allow members of the wheel group to execute the defined actions 
+ * without password authentication, similar to "sudo NOPASSWD:"
+ */
+polkit.addRule(function(action, subject) {
+    if ((action.id == "org.gameros.steamvr.pkexec") &&
+        subject.isInGroup("wheel"))
+    {
+        return polkit.Result.YES;
+    }
+});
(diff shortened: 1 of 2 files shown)


rootfs/etc/polkit-1/rules.d/41-steamvr.rules
```diff
@@ -0,0 +1,10 @@
+/* Allow members of the wheel group to execute the defined actions
```
alkazar: doesn't this allow any action?

ShadowApex: No, this should only allow users in the `wheel` group to execute the "org.gameros.steamvr.pkexec" action, which narrowly defines running setcap on vrcompositor-launcher.

alkazar: oh, yeah, my mistake.

PR #82, opened 2020-03-24
Add vkbasalt package from AUR

ShadowApex:
This PR adds the vkbasalt package from AUR. vkBasalt is a Vulkan post processing layer to enhance the visual graphics of games, which can be optionally enabled per game. More information about it is here:

https://github.com/DadSchoorse/vkBasalt

### DIFF
--- a/profiles/default
+++ b/profiles/default
@@ -91,6 +91,7 @@ export AUR_PACKAGES="\
 	lib32-gamemode \
 	mangohud \
 	lib32-mangohud \
+	vkbasalt \
 "
 
 export SERVICES="\


alkazar:
This package is failing to build for me right now.
```
install: cannot stat 'build/libvkbasalt64.so': No such file or directory
==> ERROR: A failure occurred in package().
    Aborting...
```

Even then, I am unsure about whether we should include this. Is it disabled by default? Does it have any effect on games if it is disabled? I have no experience using this.

ShadowApex:
@alkazar Yeah, I noticed the same issue. If `CXXFLAGS` is set in the environment, part of the default in `vkBasalt/src/makefile` are ignored:
https://aur.archlinux.org/packages/vkbasalt/?O=10&PP=10#comment-727397

> Even then, I am unsure about whether we should include this. Is it disabled by default? Does it have any effect on games if it is disabled? I have no experience using this.

I wanted to include it in a PR for consideration, but I'm fine if you wanted to exclude it. By default, vkbasalt is disabled unless you configure the launch options with `ENABLE_VKBASALT=1 %command%`. You can see the post-processing effects it can provide to games here: 
https://www.youtube.com/watch?v=hSlaGkbTRi8

PR #83, opened 2020-03-24
Add steam-native-runtime meta package

ShadowApex:
This PR adds the `steam-native-runtime` meta package. Without it, I have seen some `libvpau.so.1` errors in SteamVR.

### DIFF
--- a/profiles/default
+++ b/profiles/default
@@ -72,6 +72,7 @@ export PACKAGES="\
 	haveged \
 	openbox \
 	termite \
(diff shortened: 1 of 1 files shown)


alkazar:
`steam-native-runtime` shouldn't make a difference since we do not use it. However, it could be that a library that is pulled in as a dependency is required for SteamVR. I'd rather figure out what that library is then pull in all of `steam-native-runtime` and include a lot of unnecessary libraries.

Issue #63, opened 2020-01-11
use github actions to build system images


alkazar:

tommy4st:
I tried to setup a build in Docker, but this failed because of the loop mount. When using ostree, you probably don't need to loop mount anymore and it will possibly work. I tried to set up a Docker container based on the [Arch image](https://hub.docker.com/_/archlinux), where all dependencies are installed. (It's a little bit difficult because makepkg won't run as root)

```sh
export BASE="$PWD"
pacman -Suy --noconfirm
pacman -S --needed --noconfirm base-devel git btrfs-progs arch-install-scripts
cd /tmp
useradd -m someone
echo "someone ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
sudo -u someone git clone https://aur.archlinux.org/pikaur.git
cd pikaur
sudo -u someone makepkg -fsri --noconfirm
cd $BASE
```

alkazar:
I think it should be possible to do this by doing the mount on the host system and doing the rest of the build in the docker container.

PR #84, opened 2020-03-24
Add docker support for building GamerOS images

ShadowApex:
This PR adds support for building GamerOS images in Docker, which addresses some points in #63. 

To get this to work, I needed to make a few modifications to `build.sh` to support running `pikaur` as a normal user. By default, `pikaur` will run as it did before, so these changes shouldn't have any effect building outside Docker.

Also added are optional image output directory environment variables which the Docker image uses to move the image file outside the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+FROM archlinux:latest
+LABEL contributor="shadowapex@gmail.com"
+
+RUN echo -e "[multilib]\nInclude = /etc/pacman.d/mirrorlist\n" >> /etc/pacman.conf && \
+	pacman --noconfirm -Syy && \
+	pacman --noconfirm -Syu && \
+	pacman --noconfirm -S arch-install-scripts btrfs-progs pyalpm sudo && \
+	pacman --noconfirm -S --needed base-devel git && \
+	echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
+	useradd build -G wheel -m && \
+	su - build -c "git clone https://aur.archlinux.org/pikaur.git /tmp/pikaur" && \
+	su - build -c "cd /tmp/pikaur && makepkg -f" && \
+	pacman --noconfirm -U /tmp/pikaur/pikaur-*.pkg.tar.xz
+
+# Add a fake systemd-run script to workaround pikaur requirement.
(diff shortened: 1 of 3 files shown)


sharkwouter:
Oh wow, this seems to actually work. That's awesome!

I was not expecint this to be possible, because we used to run mkinitcpio in the build script. When did that change?

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM archlinux:latest
+LABEL contributor="shadowapex@gmail.com"
+
+RUN echo -e "[multilib]\nInclude = /etc/pacman.d/mirrorlist\n" >> /etc/pacman.conf && \
+	pacman --noconfirm -Syy && \
+	pacman --noconfirm -Syu && \
+	pacman --noconfirm -S arch-install-scripts btrfs-progs pyalpm sudo && \
+	pacman --noconfirm -S --needed base-devel git && \
+	echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
```
sharkwouter: Do we even need sudo? The build runs as root

build.sh
```diff
@@ -8,9 +8,11 @@ if [ $EUID -ne 0 ]; then
 	exit 1
 fi
 
-SYSTEM_DESC=GamerOS
-SYSTEM_NAME=gameros
-USERNAME=gamer
+SYSTEM_DESC=${SYSTEM_NAME:-GamerOS}
+SYSTEM_NAME=${SYSTEM_NAME:-gameros}
+USERNAME=${USERNAME:-gamer}
+BUILD_USER=${BUILD_USER:-}
```
sharkwouter: I'm not sure if this pull request should include adding support for building packages with a specified user, especially since the docker image doesn't seem to use it.

Dockerfile
ShadowApex: Yes, `sudo` is needed as pikaur runs as a non-root user. This is done because pikaur spawns package builds using `systemd-run` to build packages as non-root, but the container does not have systemd running as its init process.

build.sh
ShadowApex: @sharkwouter Allowing a different build user is essential to getting it to work in Docker. You can see that it's used in the [Dockerfile](/Dockerfile) defined as an environment variable. This is done because pikaur spawns package builds using `systemd-run` to build packages as non-root, but the container does not have systemd running as its init process.

sharkwouter:
Thanks for the explanations! I see no issues with it then

PR #90, opened 2020-04-04
Feature/automount usb

sharkwouter:
This change adds a udev rule which will automatically mount USB drives in ~/mount in their own directory. It also adds a cleanup for when the USB drive is removed, but that currently runs a bit too early to clean up the directory in which the USB drive was mounted. It will, however, clean up other empty directories in ~/mount. I will look into this and update the PR if I find a solution.


sharkwouter:
I'm going to close this for now. I'm looking into a slightly different solution at the moment.

PR #93, opened 2020-04-15
Update build.sh

liberodark:
Little fixe and this need other work for better read.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,7 +4,7 @@ set -e
 set -x
 
 if [ $EUID -ne 0 ]; then
-	echo "$(basename $0) must be run as root"
+	echo "$(basename "$0") must be run as root"
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -183,29 +183,29 @@ mkdir /var
...
-rm ${BUILD_PATH}/manifest
+echo "${CHANNEL}-${VERSION}" > "${BUILD_PATH}"/build_info
+echo "" >> "${BUILD_PATH}"/build_info
+cat "${BUILD_PATH}"/manifest >> "${BUILD_PATH}"/build_info
+rm "${BUILD_PATH}"/manifest
 
-btrfs subvolume snapshot -r ${BUILD_PATH} ${SNAP_PATH}
-btrfs send -f ${CHANNEL}-${VERSION}.img ${SNAP_PATH}
+btrfs subvolume snapshot -r "${BUILD_PATH}" "${SNAP_PATH}"
+btrfs send -f "${CHANNEL}"-"${VERSION}".img "${SNAP_PATH}"
```
alkazar, line 192: In this case I think it would look better to write `"${CHANNEL}-${VERSION}.img"`

liberodark: yeah is little same just have secure var.

PR #147, opened 2020-10-10
Add ppsspp for PSP support

sharkwouter:
This adds a patched version of the libretro-ppsspp package found in the Arch repo. The difference is that this package includes the assets and has been patched to use them. It is required for https://github.com/gamer-os/steam-buddy/pull/101

### DIFF
--- a/manifest
+++ b/manifest
@@ -117,6 +117,7 @@ export AUR_PACKAGES="\
 	alienware-alpha-wmi \
 	bcache-tools \
 	mame-tools \
+	libretro-ppsspp-gameros \
 "
 
 export SERVICES="\
--- a/manifest
+++ b/manifest
@@ -5,7 +5,7 @@ export SYSTEM_DESC="GamerOS"
 export SYSTEM_NAME="gameros"
 export USERNAME="gamer"
 export SIZE="6500MB"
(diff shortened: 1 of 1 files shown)


alkazar:
Blocked on merging PSP support into Steam Buddy.

sharkwouter:
Actually, it would be better to add this as dependency for steam-buddy. Somehow we have a lot of emulators in this file here, even though steam-buddy installs them already.

alkazar:
Good point. Not sure if I want to make `libretro-ppsspp-gameros` a dependency of `steam-buddy` though.

sharkwouter:
I changed the PR to install the package before steam-buddy and changed the package itself to provide libretro-ppsspp. That means you can add libretro-ppsspp as a dependency for steam-buddy and this package will statisfy that.

I'm not sure what is up with the build error, though. I think that is unrelated.

sharkwouter:
It seems the error is indeed completely unrelated to the changes in this PR. Game mode was added to the official repo now. Should we fix this in a different PR?

sharkwouter:
I've changed the archive version used to make sure the image can be build with gamemode from the repository, since it's no longer in the chaotic-aur.

PR #159, opened 2020-11-10
Tweak Retroarch audio latency setting

rshamsnejad:
I experienced some pops and clicks in Super Mario 64 game that went away with this setting (previously set at 64ms). This shouldn't create side effects at all.


alkazar:
This change cannot be done here. There can be no files under `/home/gamer` in the system image. The actual home directory is mounted on top.

You can make this config change in the Steam Buddy repo.

PR #168, opened 2020-12-25
Better handling of Bluetooth dongles 

rshamsnejad:
I was trying to make the following dongle work with GamerOS : 
image

And to succeed I had to do two operations : 

1. Force HCI mode to make the dongle an actual Bluetooth dongle, because it goes into HID mode by default. This can be achieved in two ways : 
- By pressing the "Connect" button while plugging it in
- By installing the `bluez-hid2hci` package from the [Arch repos](https://archlinux.org/packages/extra/x86_64/bluez-hid2hci/), in order to [automatically force it](https://wiki.archlinux.org/index.php/Bluetooth#Logitech_Bluetooth_USB_Dongle).

The second option is obviously way more convenient, so this pull request adds the package to the `manifest` file in a first commit.

2. Force the Bluetooth controller to automatically be powered on when plugged in : 
If the dongle is plugged in before boot, everything works fine. But if it is plugged after boot, it's necessary to connect through SSH and launch `$ bluetoothctl power on` before any connection can take place.

This pull requests prevents the need to do this by activating an option in a config file as described in the [Arch Wiki](https://wiki.archlinux.org/index.php/Bluetooth#Auto_power-on_after_boot) in a second commit.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -182,7 +182,7 @@ DISTRIB_DESCRIPTION=${SYSTEM_DESC}
 find /usr/share/libretro/autoconfig -type f -name '*.cfg' | xargs -d '\n' sed -i '/input_menu_toggle_btn/d'
 
 # automatically power-on Bluetooth dongles when plugged in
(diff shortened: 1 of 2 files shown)


alkazar:
I wonder if with this, we no longer need this:
https://github.com/gamer-os/steamos-compositor-plus/blob/master/usr/bin/steamos-session#L29

Will have to test it out I suppose. I added that line because my bluetooth devices didn't work without it.

alkazar:
Nice work!

PR #233, opened 2021-06-08
Avoid journald chewing up 4G of disk space

Samsagax:
On most computers with the current installation, logs made by [`journald` will chew up 4Gib](https://wiki.archlinux.org/title/Systemd/Journal#Journal_size_limit) that can be used for more interesting stuff.

### DIFF
--- a/rootfs/etc/systemd/journald.conf.d/00-journal-size.conf
+++ b/rootfs/etc/systemd/journald.conf.d/00-journal-size.conf
@@ -1,0 +1,2 @@
+[Journal]
+SystemMaxUse=50M


sharkwouter:
Oh cool, that's easy to implement. Just add the configuration file for it to rootfs. What would be a sane default? We don't need information from many boots I'd say. Maybe 1 GB or a couple of 100 MB

Samsagax:
> Oh cool, that's easy to implement. Just add the configuration file for it to rootfs. What would be a sane default? We don't need information from many boots I'd say. Maybe 1 GB or a couple of 100 MB

I went for the suggested in the ArchWiki. And in my experience is with 50Mb is more than enough for diagnosing almost any trouble. It can contain the couple reboots prior and the current session. Is about 250k journal lines.

sharkwouter:
Makes sense then. The 50 MB seemed a bit arbitrary to me initially. I won't be able to test it unfortunately

Samsagax:
I think you can test the effect on any Linux box that runs systemd. Just check the size of `/var/log` before and after the change.

PR #244, opened 2021-07-06
Enable Kernel Modesetting for `nvidia`

Samsagax:
This is needed for Wayland support, Prime support and some setups may benefit from it even if we don't support them.


alkazar:
Do we also need to add support for this to the initramfs?

Samsagax:
> Do we also need to add support for this to the initramfs?

Only is needed if we need to support early KMS. Most setups should work fine without it. But maybe we can add all GPU drivers to the initramfs the kernel will use the ones that need but if you do that, you need to add the conf file to the `FILES` array in `mkinicpio.conf` file for the modules to pick any options you may set.

PR #258, opened 2021-07-29
Increase size of builder image

Samsagax:


Samsagax:
Closed in favor of #259

PR #263, opened 2021-08-11
Add os-release in rootfs

panos:
First contribution in a long while.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -176,6 +176,18 @@ DISTRIB_RELEASE=\"${LSB_VERSION}\"
 DISTRIB_DESCRIPTION=${SYSTEM_DESC}
 " > /etc/lsb-release
 
+echo '
+NAME="${SYSTEM_DESC}"
+VERSION="${VERSION}"
+PRETTY_NAME="${SYSTEM_DESC} ${VERSION}"
+ID=${SYSTEM_NAME}
+ID_LIKE=arch
+ANSI_COLOR="1;31"
+HOME_URL="${WEBSITE}"
+DOCUMENTATION_URL="${DOCUMENTATION_URL}"
+BUG_REPORT_URL="${BUG_REPORT_URL}"
+' > /etc/os-release
+
 # install extra certificates
 trust anchor --store /extra_certs/*.crt
 
--- a/manifest
+++ b/manifest
@@ -6,6 +6,9 @@ export SYSTEM_NAME="chimeraos"
 export USERNAME="gamer"
 export SIZE="10000MB"
 export ARCHIVE_DATE=""
+export WEBSITE="https://chimeraos.org"
+export DOCUMENTATION_URL="https://chimeraos.org/about"
+export BUG_REPORT_URL="https://github.com/ChimeraOS/chimeraos/issues"
 
 export PACKAGES="\
 	amd-ucode \


panos:
To be kept in mind, by the way: https://chimera-linux.org/

rootfs/etc/os-release
```diff
@@ -0,0 +1,9 @@
+NAME="Chimera OS"
+PRETTY_NAME="Chimera"
+ID=chimera
+ID_LIKE=arch
+BUILD_ID=rolling
```
Samsagax, line 5: We are not exactly a rolling release. We ship snapshots every month or so.

rootfs/etc/os-release
```diff
@@ -0,0 +1,9 @@
+NAME="Chimera OS"
+PRETTY_NAME="Chimera"
```
Samsagax, line 2: Should include the release (you can get it as we tag each pre-release with the number and commit). Maybe you can do a @@ substitution in the build process.

rootfs/etc/os-release
```diff
@@ -0,0 +1,9 @@
...
+ID=chimera
+ID_LIKE=arch
+BUILD_ID=rolling
+ANSI_COLOR="1;31"
+HOME_URL="https://chimeraos.org/"
+DOCUMENTATION_URL="https://chimeraos.org/about"
```
Samsagax, line 8: I think this should point to the wiki.

rootfs/etc/os-release
```diff
@@ -0,0 +1,9 @@
...
+BUG_REPORT_URL="https://github.com/ChimeraOS"
```
Samsagax, line 9: This should point to the github issues page.

Samsagax (changes requested):

panos:
Keep in mind that many of the links here are the links that the website itself pointed to, but you're absolutely correct. I'll ship the new changes later. @Samsagax

rootfs/etc/os-release
alkazar: The wiki is being phased out, so I think for now the about page is okay, or we can omit it.

rootfs/etc/os-release
```diff
@@ -0,0 +1,9 @@
+NAME="Chimera OS"
```
alkazar, line 1: NAME and PRETTY_NAME should probably just be "ChimeraOS"

rootfs/etc/os-release
Samsagax: We should have some kind of documentation with common Troubleshooting at least. Maybe not in a GitHub wiki, but in a markdown file somewhere.

rootfs/etc/os-release
panos: Arch Linux uses Arch Linux in both fields.

Samsagax: > Arch Linux uses Arch Linux in both fields.

Yeah, that makes sense since they don't have named releases. I'm just saying this could help us on identifying the specific release. But If there is no objection, using `ChimeraOS` for both should be fine

alkazar: From the definition it sounds like @Samsagax is right and `NAME` should be `ChimeraOS` and `PRETTY_NAME` should be `ChimeraOS <RELEASE NUMBER>`

panos: Someone will have to remember to replace the release number every time. Should I set the number to 27?

Samsagax: You need to modify the build.sh script to replace a variable inside the file at build time. Maybe the best way is to plug that on the postinstallhook but I would prefer if the creation is done at the same time as the replacement.

alkazar: I think the easiest and best thing to do is to generate the file in `build.sh` similar to how we generate `/etc/lsb_release`.
See here: https://github.com/ChimeraOS/chimeraos/blob/master/build.sh#L172

panos:
I haven't forgotten about this PR yet, will finish it up based on your feedback.

panos:
Fixed. `BUILD_ID` was removed because it was optional anyways and only useful for differentiating between builds of, say, the same version but with a different ID. I added an extra entry for the version itself, added the version to the `PRETTY_NAME` field and put everything in `build.sh`. I used single quotes rather than double quotes.

Hope I didn't miss anything.

build.sh
```diff
@@ -176,6 +176,18 @@ DISTRIB_RELEASE=\"${LSB_VERSION}\"
 DISTRIB_DESCRIPTION=${SYSTEM_DESC}
 " > /etc/lsb-release
 
+echo '
+NAME="ChimeraOS"
```
alkazar, line 180: The build script tries to be agnostic to make it easy to edit everything about the OS in one place and allow anyone to make a fork with a new name easily.

All these values should be defined in the `manifest` file.

For example `$SYSTEM_DESC` is already defined as "ChimeraOS".
And we could use `$SYSTEM_NAME` for `ID`.
The other values will need to be added.

panos:
Not sure if I did this correctly this time around.

alkazar:
Looks good, thanks!

PR #286, opened 2021-12-13
Create udev rule to ignore asrock led controller

garrlker:
This PR creates a udev rule that ignores Asrock's motherboard rgb controller.
This is to prevent retroarch from using it as a gamepad input so me (and others) can play our retro games.


Samsagax:
If you can confirm the controller itself works for it's intended purpose (other than been seen as a joypad) seems good to me.

garrlker:
@Samsagax 

You were right, the original commit prevented using the rgb controller for its intended purpose.
This latest change just removes the js and event interfaces so retroarch doesn't detect it as a game controller.

On my own machine I can play retroarch while using a cli tool via ssh to update my rgb colors effects.

alkazar:
Awesome, thanks for the submission!

Issue #297, opened 2022-03-03
Implementation of quick suspend / resume


marcelschoen:
I heard that Valve implemented a proper quick suspend / resume feature in SteamOS. This is a big issue for a mobile device like the Aya Neo. I have currently ChimeraOS running on it, and it's great, but this one issue is a real problem when your game doesn't allow you to quicksave at any given moment. Not to mention that after switching the Neo off you'll have to restart the game from scratch. 

So my question is if it's feasible, and possibly even planned, to bring such a feature to ChimeraOS?

pastaq:
Issue resolved by #300

alkazar:
Suspend/resume when pushing the power button was enabled in ChimeraOS 32.

PR #298, opened 2022-03-05
Fix Aya Neo Suspend

pastaq:
All Aya Neo 2021 and 2021 Pro models use the mt7921e kernel module for wifi. This kernel module crashes on suspend, breaking wifi and slowing the system as it tries repeatedly to reinitialize the driver. This solution resolves that issue in a way that will enable additional failed kernel modules to easily be added in the future if discovered.
- Added systemd-suspend-modules to AUR packages manifest.
- Added /etc/suspend-modules.conf configured for mt7921e kernel module.


pastaq:
First step is solving #297

alkazar:
This is great, thanks!

PR #299, opened 2022-03-06
Disable Systemd handling of XF86PowerOff events

pastaq:
Adds a logind.conf.d folder with a config file that ignores power button pressing. Prerequisite for https://github.com/ChimeraOS/steamos-compositor-plus/pull/48 to be enabled.


pastaq:
Branch has some issues with previous PR. Reopening.

PR #303, opened 2022-03-10
Fix kernel panic on Aya Neo 2021

pastaq:
Unload mt7921e driver on shutdown to prevent bug found here https://discordapp.com/channels/560281725175988233/560281725175988238/951312146963247155


rootfs/usr/lib/systemd/system-shutdown/mt7921e.shutdown
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+# Unload the mt7921e driver to avoid kernel panic on reboot.
+rmmod mt7921e
```
alkazar, line 3: Should it be `/usr/bin/rmmod` ? At least that is what I see (full paths) being used in other scripts.

pastaq: Good catch. I used the sbin path as it requires root privileges.

PR #304, opened 2022-03-17
Makefile

pastaq:
Added a makefile to simplify building images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
 ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
-CHANNEL=
-VERSION=
+ifndef CHANNEL
+$(error CHANNEL is not set. Must set CHANNEL=value in the build command.)
+endif
+ifndef VERSION
+$(error VERSION is not set. Must set VERSION=value in the build command.)
+endif
 
 .PHONY: build
 build: builder
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,9 @@
+.PHONY: build
+build: builder
+	docker run --rm -v $(PWD)/output:/output --privileged=true image-builder:latest
+
+.PHONY: builder
+builder:
+	docker build -t image-builder:latest .
+
+
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 .PHONY: build
 build: builder
+	docker pull archlinux:base-devel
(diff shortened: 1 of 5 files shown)


pastaq:
I identified a bug that I need to fix. Currently the $(PWD) isn't functional so it will mount a directory at the root level of the host OS rather than create an output directory inside the current working directory.

Makefile
```diff
@@ -0,0 +1,19 @@
+ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+
+ifndef CHANNEL
```
alkazar, line 3: Actually, CHANNEL is not used anymore. The documentation is outdated. Sorry about that.

manifest
```diff
@@ -1,6 +1,5 @@
 #! /bin/bash
 
-export VERSION="32"
```
alkazar, line 3: I would prefer to keep the VERSION in this file. The manifest file is meant to hold all the details of the system image. Also, without it, the automated build would not work.

Makefile
```diff
@@ -0,0 +1,19 @@
...
+$(error CHANNEL is not set. Must set CHANNEL=value in the build command.)
+endif
+ifndef VERSION
+$(error VERSION is not set. Must set VERSION=value in the build command.)
+endif
+
+.PHONY: build
+build: builder
+	docker pull archlinux:base-devel
+	docker run --rm -v $(ROOT_DIR)/output:/output --privileged=true image-builder:latest $(CHANNEL) $(VERSION)
```
alkazar, line 13: The command needs to include the git commit hash instead of CHANNEL and VERSION

This is actually my local makefile:
```
system_image: build_image
        docker run --rm -v $(pwd)/output:/output --privileged=true image-builder:latest $(git rev-parse HEAD | cut -c1-7)

build_image:
        docker build -t image-builder:latest .
```

alkazar:
Just in case you are not aware, there is a GitHub action to build the system image. It is usually much easier to let GitHub to do the build. It will also update the image as well. You can even point your ChimeraOS install to your own fork by modifiying `/frzr_root/source` appropriately.

The GitHub actions are defined here:
https://github.com/ChimeraOS/chimeraos/blob/master/.github/workflows/main.yml

pastaq:
Interesting. I was not aware of that.

pastaq:
Based on the discussion I believe this to be unnecessary and less intuitive than the git way of doing things so I'm going to close it for now.

PR #309, opened 2022-04-07
Create 99-hidapi.rules for Nintendo Joy-Con and Pro Controller

cwalker:
This will give read and write permissions to hidraw-devices
Such as Nintendo Joy-Con and Pro Controller

### DIFF
--- a/rootfs/etc/udev/rules.d/99-hidapi.rules
+++ b/rootfs/etc/udev/rules.d/99-hidapi.rules
@@ -1,0 +1,2 @@
+#This will give read and write permissions to hidraw-devices
+KERNEL=="hidraw*", SUBSYSTEM=="hidraw", GROUP="1000", MODE="0664"


alkazar:
I was just doing some research on this and it seems like `hid-hidraw` and `hid-nintendo` conflict.

Shouldn't we be using `hid-nintendo` ?

cwalker:
Yes, hid-nintendo works flawesly, except gyro.
With this rule, gyro works.
As variant, just add this aur package: https://aur.archlinux.org/packages/nintendo-udev

Samsagax:
If already is an AUR package I would go for that route. The `nintendo-udev` package rule seems well-written and very device specific so looks good to me

alkazar:
Closed in favor of https://github.com/ChimeraOS/chimeraos/pull/325

Thanks for bringing this to our attention!

PR #326, opened 2022-04-13
Only trigger build on commit to master

Samsagax:
Only trigger the image generation if a push to master. Since every other branch should not create a release and ideally would be tested locally.


alkazar:
This should be much better, thanks!

PR #327, opened 2022-04-13
Aya Neo Fixes

pastaq:
Install aya-neo-fixes-git during image build.
- Replaces various fixes for the Aya Neo with a dedicated package
- Adds fix for extended button functionality for all models.
- Removes dependency for systemd-suspend-modules.


manifest
```diff
@@ -185,6 +186,7 @@ export SERVICES="\
 	haveged \
 	systemd-timesyncd \
 	sshd \
+	neo-controller \
```
alkazar, line 189: Does this service run on all hardware? Could that cause problems?

pastaq: The script the service executes checks for known Aya neo devices first and exits if it doesn't exist.

https://github.com/ShadowBlip/aya-neo-fixes/blob/c5b7293813415b2cb6245079597060efb7b65648/neo-controller.py#L27

PR #333, opened 2022-05-06
format-media.sh sudoers

pastaq:
requirement for https://github.com/ChimeraOS/chimera/pull/230


alkazar:
Have you tested this? Specifically, I am not sure that you will be able to pass parameters to `format-media.sh`.
From what I recall of the sudoers file, you have to specify the exact command including parameters which may not be possible in this case.

pastaq:
It seems you are correct. My initial test was flawed, and retained the sudo auth from editing visudo. I will continue to look for a way to solve this as wildcards don't seem to be working as documented either.

alkazar:
Awesome, thanks!

PR #334, opened 2022-05-07
Phantom Input Remover

pastaq:
Found a more thorough workaround for the phantom controller input. Enables service from [here](https://github.com/ShadowBlip/aya-neo-fixes/blob/main/phantom-input.service).

### DIFF
--- a/manifest
+++ b/manifest
@@ -188,6 +188,7 @@ export SERVICES="\
 	systemd-timesyncd \
 	sshd \
 	neo-controller \
+	phantom-input \
 "
 
 export USER_SERVICES="\


alkazar:
Couldn't this cause issues? We should probably do this from the same service and only apply it if on the Aya Neo.

pastaq:
https://github.com/ShadowBlip/aya-neo-fixes/commit/7af92b09f34d89cc2652d4a3f2043a2ace3aa80b
I made the new service dependent on neo-controller.service. Tested on my laptop that it will stop the service if neo-controller exits.

I had to make this it's own service because starting a new thread for the run loop in the neo-controller.py script was having strange behaviors with asyncio, which broke the controller.

PR #341, opened 2022-05-18
Use mangohud-x11 from AUR

Samsagax:
The AUR package has been split into Wayland and X11 (XWayland) version. The `-x11` variant includes mangoapp binaries.


alkazar:
Does the -x11 variant work in gamescope?

Samsagax:
> Does the -x11 variant work in gamescope?

Yes. Tested and works with the latest changes. You can control it within gamepadui. The binary that was missing were `mangoapp`

PR #350, opened 2022-07-05
Fix delay on startup for some users, sudoers.

pastaq:
- Fix delay on startup for some users by adding dmidecode package.
- Add /etc/sudoers.d/steam for steam client root requests. 
- Add /etc/sudoers.d/chimera for chimera root requests.
- Remove overwriting sudoers file in build.sh

Normally I would have done this in 2 PR, but considering I had to edit the sudoers anyway I felt improving the practice would be better. Technically it would be better for the chimera app and steam-removable-media to include their own sudoers files, so I can remove them from this PR and add them to their respective packages if preferred. Considering others may be using chimera I thought it warranted discussion first before changing something that significant.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -128,6 +128,14 @@ passwd --lock root
 groupadd -r autologin
 useradd -m ${USERNAME} -G autologin,wheel
 echo "${USERNAME}:${USERNAME}" | chpasswd
+echo "${USERNAME} ALL=(ALL) NOPASSWD: /usr/bin/steamos-update
+${USERNAME} ALL=(ALL) NOPASSWD: /usr/bin/jupiter-biosupdate
+${USERNAME} ALL=(ALL) NOPASSWD: /usr/bin/dmidecode -t 11
+" > /etc/sudoers.d/steam
+echo "${USERNAME} ALL=(ALL) NOPASSWD: /usr/bin/chimera-session-use-gamescope
+${USERNAME} ALL=(ALL) NOPASSWD: /usr/bin/chimera-session-use-lightdm
+${USERNAME} ALL=(ALL) NOPASSWD: /usr/lib/media-support/format-media.sh*
+" > /etc/sudoers.d/chimera
 
 # set the default editor, so visudo works
 echo "export EDITOR=/usr/bin/vim" >> /etc/bash.bashrc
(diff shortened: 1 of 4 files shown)


Samsagax:
LGTM

alkazar:
I would prefer if we don't hard code the `gamer` user.

alkazar:
In my testing I found that parameters don't need to be specified. Is this not the case?

pastaq:
> In my testing I found that parameters don't need to be specified. Is this not the case?

It is better practice to specify params when possible as it prevents arbitary code execution. clever use of back-tick for example can allow users to navigate to root using \`../../` and execute any command as root.

pastaq:
> I would prefer if we don't hard code the `gamer` user.

Alright. Then the only real option is to include them in the packages and use $USER/$SUDO_USER, or change the build script back but have it create these files. Steam will probably need to happen that way.

Alternatively, we can investigate polkit handlers instead. That is how SteamOS is doing it now, but it may essentially be and ALL ALL=(ALL) permission.

Samsagax:
The polkit solution seems cleaner to me but until we figure that out it would be better to have this in separate files under `/etc/sudoers.d`, I think is better than just mes up with sodoers file directly. I tested those and they work as intended with the dummy scripts we already provide.

pastaq:
> The polkit solution seems cleaner to me but until we figure that out it would be better to have this in separate files under `/etc/sudoers.d`, I think is better than just mes up with sodoers file directly. I tested those and they work as intended with the dummy scripts we already provide.

Alright, so the way forward I think will be to edit the build file for now to create the distinct sudoers.d files. I will open an issue on each of the respective projects to handle thier own sudoers during install. The steam permissions on the ChimeraOS repo, chimera session scripts on the Chimera repo, and steam removable media on its repo. Then each change can be staged and we can trigger the merges when everyone is synced up. Does everyone concur?

build.sh
```diff
@@ -128,16 +128,14 @@ passwd --lock root
 groupadd -r autologin
 useradd -m ${USERNAME} -G autologin,wheel
 echo "${USERNAME}:${USERNAME}" | chpasswd
-echo "
-root ALL=(ALL) ALL
-${USERNAME} ALL=(ALL) ALL
```
alkazar, line 133: does removing this mean the user cannot use `sudo`? Or is there a default file?

pastaq: By default it should already include a sudoers file. I haven't built an image to verify...

alkazar: Okay, well, I guess we'll find out.

PR #352, opened 2022-07-09
Remove deprecated aya-neo-fixes package, include replacements.

pastaq:


pastaq:
As an aside, we should wait for https://github.com/libretro/retroarch-joypad-autoconfig/pull/848 before building, as the retroarch config needs to be added.

PR #359, opened 2022-07-22
Make workdir a mandatory docker volume

Samsagax:
Image wont build if no volume is mounted for `/workdir` thus making less likely to use wrong build files.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
         run: docker build --no-cache -t image-builder:latest .
       - name: Build system image
         id: build_image
-        run: docker run --rm -v $(pwd)/output:/output --privileged=true image-builder:latest $(echo ${GITHUB_SHA} | cut -c1-7)
+        run: docker run --rm -v $(pwd):/workdir -v $(pwd)/output:/output --privileged=true image-builder:latest $(echo ${GITHUB_SHA} | cut -c1-7)
       - name: Create release
         id: create_release
         uses: actions/create-release@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         run: docker build --no-cache -t image-builder:latest .
       - name: Build system image
         id: build_image
-        run: docker run --rm -v $(pwd)/output:/output --privileged=true image-builder:latest $(echo ${GITHUB_SHA} | cut -c1-7)
+        run: docker run --rm -v $(pwd):/workdir -v $(pwd)/output:/output --privileged=true image-builder:latest $(echo ${GITHUB_SHA} | cut -c1-7)
```
alkazar, line 20: does this actually work? I recently noticed that `$(pwd)` doesn't work locally and I had to specify the full path manually.

Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN echo -e "#!/bin/bash\nif [[ \"$1\" == \"--version\" ]]; then echo 'fake 244
 RUN reflector --verbose --latest 20 --sort rate --save /etc/pacman.d/mirrorlist
 
 # Add the project to the container.
-ADD . /workdir
+#ADD . /workdir
```
alkazar, line 27: May as well delete this and the line above

.github/workflows/main.yml
Samsagax: I use the exact same line as here and works for me. By any chance did you set pwd as a variable or are using any other shell? (I'm using bash)

Dockerfile
Samsagax: Totally.

PR #362, opened 2022-07-23
Add bluetooth-workaround service file

Samsagax:
Bluetooth wont start if it doesnt have `/var/lib/bluetooth` directory in
place. Since we overlay `/var` we cant deploy the directory Archlinux
worked around in their package so we add a tiny service that will just
create that directory.

### DIFF
--- a/rootfs/etc/systemd/system/bluetooth-workaround.service
+++ b/rootfs/etc/systemd/system/bluetooth-workaround.service
@@ -1,0 +1,10 @@
+[Unit]
+Description="Bluetooth workaround"
+
+[Service]
+Type=oneshot
+ExecStart=mkdir -p /var/lib/bluetooth
+Before=bluetooth.service
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


rootfs/etc/systemd/system/bluetooth-workaround.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description="Bluetooth workaround"
+
+[Service]
+Type=oneshot
+ExecStart=mkdir -p /var/lib/bluetooth
+Before=bluetooth.service
+
+[Install]
+WantedBy=multi-user.target
```
alkazar, line 10: does this ensure that this will run before the bluetooth service is started?

Samsagax: Yes. Exactly. The "Before" will wait for this to finish before starting bluetooth service. I even destroyed the directory on purpose and this will fix it.

alkazar: I am blind and didn't notice that line, thanks. :facepalm:

PR #366, opened 2022-07-28
Add xorg-xdpyinfo

flexiondotorg:
Add `xorg-xdpyinfo` to the PACKAGES manifest so that the proposed changes to system-info (https://github.com/ChimeraOS/chimera/pull/236) closely match what Steam System Information generates.

### DIFF
--- a/manifest
+++ b/manifest
@@ -17,6 +17,7 @@ export PACKAGES="\
 	lightdm \
 	accountsservice \
 	xorg-server \
+	xorg-xdpyinfo \
 	bluez \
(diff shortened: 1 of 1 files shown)


Samsagax:
As pointed out in [this comment](https://github.com/ChimeraOS/chimera/pull/236#discussion_r932145343) this should be a dependency of chimera if we chose to do so.

alkazar:
Agreed with @Samsagax I will add the dependency to the chimera app's PKGBUILD with the next release.

PR #368, opened 2022-07-28
Add os-release info for display

Samsagax:
This add missing values to `/etc/os-release` so GamepadUI will pick them up.


alkazar:
I don't think we need or want that. It adds no additional useful information.

alkazar:
I don't think we should put inaccurate information into those fields just to "make it look better".

pastaq:
We could use this to display the actual install channel org/repo/tag could go there.

Samsagax:
> I don't think we should put inaccurate information into those fields just to "make it look better".

I wouldn't call it inaccurate. But fair enough. Could be useful though.
@pastaq the tag is already there, the problem with adding the update channel is that the image is not modified, the release on GitHub is marked as release.

PR #369, opened 2022-07-31
Session switcher rework to use LightDM for all sessions

Samsagax:
Split LightDM configuration files into a basic configuration for all sessions, a user autologin name and a session declaration. Rework chimeraos-session-switcher to use 10-chimeraos-session.conf to change sessions. Tested on my branch and works.

Further work would need to add steamos-session-switcher to be able to work on Steam client (at leas on gamepadui) and remove some leftovers.


alkazar:
Nice!

PR #373, opened 2022-08-03
Use gamescope-plus

Samsagax:
This adds `--force-orientation ` for use on handhelds or any other device that needs screen rotation.


ruineka:
This patches gamescope stable instead of the latest git so as a bonus this may solve the issues with older AMD GPUs as well. @alkazar

PR #378, opened 2022-08-12
Desktop stuff

Samsagax:
Some desktop defaults and notable changes:
- Remove `gnome-desktop` as it is a dependency of other packages (and most are starting to migrate to `gnome-desktop-4`)
- Add some basic apps for a functional desktop (text editor, archiver, image viewer, document reader, thumbnailer, system monitor)
- Add flatpak support (and fix slowdowns in app launchs)
- Add user directories support
- Add branded background and set it as default and prefer dark theme for all GTK 3 and 4 apps.


rootfs/etc/dconf/db/local.d/01-dark-theme
```diff
@@ -0,0 +1,3 @@
+[org/gnome/desktop/interface]
```
alkazar, line 1: Dark mode is ugly an inconsistent. Let's not turn it on by default.

rootfs/etc/dconf/profile/user
```diff
@@ -0,0 +1,2 @@
+user-db:user
```
alkazar, line 1: what is this for?

manifest
```diff
@@ -11,13 +11,22 @@ export DOCUMENTATION_URL="https://chimeraos.org/about"
...
 
 export PACKAGES="\
-	gnome-desktop \
 	gnome-control-center \
 	gnome-session \
 	gnome-shell \
 	gnome-software \
 	gnome-terminal \
+	gnome-keyring \
+	gnome-menus \
```
alkazar, line 20: What does this do?

manifest
```diff
@@ -11,13 +11,22 @@ export DOCUMENTATION_URL="https://chimeraos.org/about"
...
+	gnome-system-monitor \
```
alkazar, line 21: I don't think we need this.

manifest
```diff
@@ -11,13 +11,22 @@ export DOCUMENTATION_URL="https://chimeraos.org/about"
...
 	nautilus \
+	sushi \
+	eog \
+	evince \
```
alkazar, line 25: I don't think we need this.

manifest
```diff
@@ -11,13 +11,22 @@ export DOCUMENTATION_URL="https://chimeraos.org/about"
...
+	file-roller \
+	gedit \
```
alkazar, line 27: Not sure about this one... you can grab any text editor from flathub.

manifest
```diff
@@ -11,13 +11,22 @@ export DOCUMENTATION_URL="https://chimeraos.org/about"
...
+	xdg-user-dirs-gtk \
```
alkazar, line 28: These directories seem rather useless to me.

manifest
```diff
@@ -11,13 +11,22 @@ export DOCUMENTATION_URL="https://chimeraos.org/about"
 export BUG_REPORT_URL="https://github.com/ChimeraOS/chimeraos/issues"
 
 export PACKAGES="\
-	gnome-desktop \
 	gnome-control-center \
 	gnome-session \
 	gnome-shell \
 	gnome-software \
 	gnome-terminal \
+	gnome-keyring \
```
alkazar, line 19: I am not certain what this is needed for.

alkazar:
I think this background looks better:

chimeraos_background2

manifest
```diff
@@ -11,13 +11,22 @@ export DOCUMENTATION_URL="https://chimeraos.org/about"
...
+	gnome-menus \
+	gnome-system-monitor \
 	nautilus \
+	sushi \
```
alkazar, line 23: Thumbnails already appear to work, how is this different/better than the default?

rootfs/etc/dconf/profile/user
Samsagax: This is for setting default configurations systemwide for all users in the system. Like this:
https://help.gnome.org/admin/system-admin-guide/stable/dconf-custom-defaults.html.en
Note we are not locking them out, it can be changed. But sets the defaults in the local db.

rootfs/etc/dconf/db/local.d/01-dark-theme
Samsagax: It's inconsistent if not set on all the gym settings. I'm not only setting it here but for GTK 3.0 also. I prefer dark mode for OLED screens but if you think is not good, we can just revert it.

manifest
Samsagax: Gnome keyring allows for password storage in the desktop. Gnome Menus is used on the "Classic" flavor of Gnome. But can be do without.

manifest
Samsagax: It can be useful for some troubleshooting tasks. Of course is just a graphical htop, but may be of assistance.

manifest
Samsagax: The primary function of the desktop would be doing some troubleshooting and configuration tasks the user can't do with the BPM or GamepadUI. Those mostly involve editing files, looking some documents and maybe images. The thumbnailer is just so it doesn't look like Windows 95.

manifest
Samsagax: True. And gedit is there in flathub. But having it native doesn't add too much and is better to have it readily available than the user needing to surf the web or the software center for it. I can foresee some users trying to install software downloading from Softnic or UptoDown because of this.
The goal is to provide a minimal functional desktop with the tools to follow a tutorial on protondb or the internet devised for the Deck (mostly).

manifest
Samsagax: They add to the desktop experience. Gnome and some flathub apps would rely on the default xdg directories to be there for downloads, images and documents. This only provides a mean for the desktop to manage them.

manifest
Samsagax: Password storage. Nothing more. It is helpful for any browser usage (although we don't provide any, flathub is at the ready). It can also help on locking ssh keys if needed.

manifest
Samsagax: We could do without it. It is also the thumbnailer if I'm correct.

alkazar: I don't think so. Thumbnails work on the current build without sushi.

manifest
alkazar: I don't think anyone will need a PDF viewer, which is what evince is.

manifest
alkazar: Well, I guess we can keep it.

manifest
alkazar: Let's skip gnome-menus then.

rootfs/etc/dconf/db/local.d/01-dark-theme
alkazar: Well, for example, I just tried installing gedit via flatpak and it doesn't honor the dark mode setting. I think we leave the default light mode to keep consistency with most apps.

manifest
alkazar: > > I think this background looks better:
> > chimeraos_background2
> 
> I do too. But since we are mostly be used on OLED or IPS displays I would like to minimize power consumption.

I don't think that is a huge issue. What I don't like about the black background is it is a usability issue because I cannot see the top menu bar.

manifest
Samsagax: Ok.

rootfs/etc/dconf/db/local.d/01-dark-theme
Samsagax: Do you have xdg-portal-gnome installed?

manifest
Samsagax: Dropped

manifest
Samsagax: But is great to not see the top menu bar... I like how it blends. Let's do this: add this one too and set as default and have the full black one as an option to the user

rootfs/etc/dconf/db/local.d/01-dark-theme
alkazar: No, is that needed?

alkazar:
Okay, good enough. Merging.

rootfs/etc/dconf/db/local.d/01-dark-theme
Samsagax: I think we would need to set `GTK_THEME_VARIANT=dark` globally for it to work.

PR #380, opened 2022-08-16
Remove bcache-tools

Samsagax:
We don't use it. If the image needs it, we should add it later.


alkazar:
I think there were some users that use this. It is not a huge package so I think it is fine to leave it in.

Samsagax:
I was experimenting with it in the past when multidisk was an issue. Now it is solved I don't see the point. But if there are some users in the wild with it I don't know how they managed to use it since we don't provide the means for such installation.

PR #381, opened 2022-08-17
Gamepadui desktop switch

Samsagax:
Basic switch that works on GamepadUI via the "Switch to Desktop" option in Power Menu. The "Return to Game mode" is handled by a desktop icon and will come back to GamepadUI.


rootfs/usr/share/applications/org.chimeraos.Gamescope.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Name=Return to Gaming Mode
+Exec=/usr/bin/steamos-session-select gamescope
+Icon=/usr/share/icons/chimeraos/scalable/actions/return-to-game-mode.svg
```
alkazar, line 4: I think using the Steam icon is probably better?

rootfs/usr/bin/steamos-session-select
```diff
@@ -0,0 +1,70 @@
...
+session="${1:-gamescope}"
+
+session_launcher=""
+create_sentinel=""
+
+
+# We use "plasma" as "desktop" to hook up to SteamOS's scripts
+case "$session" in
+  plasma)
+    session_launcher="gnome-wayland"
```
alkazar, line 22: When we run `chimera-session desktop` I believe we currently use gnome X11. Should that be unified?
I also recall having trouble getting the gnome wayland to work properly.

rootfs/usr/share/applications/org.chimeraos.Gamescope.desktop
Samsagax: We need to add it. Can be done. But maybe we should have a broader discussion about branding the OS?

rootfs/usr/bin/steamos-session-select
Samsagax: Wayland depends mostly on driver support... Nvidia could have a bad time(tm) but it mostly works for me on my laptop. The issuel with X11 is it doenst work well with touchscreens. I figured if gamescope session works, gnome-wayland should definitely work. Remember this switcher is only meant to be for gamepadui and back but it can be adapted to work based on a sessing on which session to use based on a config file (didn't do it because I'm lazy :P ).

alkazar: Ah, good point.

rootfs/usr/share/applications/org.chimeraos.Gamescope.desktop
alkazar: I'll merge now, we can figure out the icon later. It should be changed to something else in any case, it is too big/detailed to be used as an icon IMO.

Samsagax: > I'll merge now, we can figure out the icon later. It should be changed to something else in any case, it is too big/detailed to be used as an icon IMO.

True that. Maybe we can do a simplified version of the lion head to be iconified (on small screens it doesn't look right)

PR #383, opened 2022-08-17
Replace pulseaudio with pipewire stack

Samsagax:
Pipewire is stable enough now to be used. Being using it for a long time on all my setups without issues.

### DIFF
--- a/manifest
+++ b/manifest
@@ -40,16 +40,16 @@ export PACKAGES="\
 	networkmanager \
 	pipewire \
 	lib32-pipewire \
+	pipewire-alsa \
+	pipewire-pulse \
+	pipewire-jack \
+	gst-plugin-pipewire \
 	xdg-desktop-portal \
 	xdg-desktop-portal-wlr \
-	pulseaudio \
-	lib32-libpulse \
-	pulseaudio-alsa \
 	alsa-utils \
 	alsa-firmware \
 	sof-firmware \
 	alsa-ucm-conf \
-	pulseaudio-bluetooth \
 	sudo \
 	python \
 	flatpak \


alkazar:
Pipewire does not work for me when plugging/unplugging headphones. Let's give it a shot on a clean system though.

manifest
```diff
@@ -40,16 +40,16 @@ export PACKAGES="\
 	networkmanager \
 	pipewire \
 	lib32-pipewire \
+	pipewire-alsa \
+	pipewire-pulse \
+	pipewire-jack \
+	gst-plugin-pipewire \
```
alkazar, line 46: What is this for?

manifest
```diff
@@ -40,16 +40,16 @@ export PACKAGES="\
 	networkmanager \
 	pipewire \
 	lib32-pipewire \
+	pipewire-alsa \
+	pipewire-pulse \
+	pipewire-jack \
```
alkazar, line 45: I don't think we need jack.

manifest
```diff
@@ -40,16 +40,16 @@ export PACKAGES="\
 	networkmanager \
 	pipewire \
 	lib32-pipewire \
+	pipewire-alsa \
```
alkazar, line 43: Should we also install wireplumber?

manifest
Samsagax: The gst-plugin is needed for streaming/recording gamescope session, gnome-shell or any other pipewire stream. It only adds 46Kb to the compressed image and 127Kb fully installed

manifest
Samsagax: I've seen some packages depending on jack, but if you see it interfering, we can drop it.

manifest
Samsagax: Wireplumber is installed as dependency of any of pipewire-{alsa,pulse,jack} packages.

PR #384, opened 2022-08-17
Introduce KERNEL_PACKAGE variable to swap kernels

Samsagax:
We dont build the kernel package because it will take too much time. Any
package we could put in that variable should be accessible via pacman
repos. Pacman configuration should be edited accordingly if non-standard
kernel is desired.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -111,6 +111,9 @@ Include = /etc/pacman.d/chaotic-mirrorlist
 # update package databases
 pacman --noconfirm -Syy
 
+# install kernel package
+pacman --noconfirm -S "${KERNEL_PACKAGE}" "${KERNEL_PACKAGE}-headers"
+
 # install packages
 pacman --noconfirm -S --overwrite '*' ${PACKAGES}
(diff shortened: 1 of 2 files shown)


alkazar:
What purpose does this serve? Not sure I like these changes, in particular switching to nvidia-dkms.

Samsagax:
It will allow for use of any kernel that is reachable by pacman. For instance, we could swith to -zen kernel, -lts kernel, -hardened kernel or even -mainline kernel (if we add the miffe repo to pacman). So in case of a cathastrophe like the current 5.19.1 that bootloops on OXP for example, we could swith to -lts. Or in case we feel like experimenting, we can go for -zen.
This will only allow for switching kernel at build time. Is not meant to be used as a mean to hot-swap hernels each boot. We are still going to replace default vmlinuz-linux image and initramfs.
As for the nvidia-dkms, I don't see any downsides. We already use a lot of dkms packages because of off-tree drivers. I'm using it myself on my laptop daily and doesnt cause any troubles.

PR #399, opened 2022-08-24
Allow removing dkms and kernel headers

Samsagax:
Allows for a slimmer image. Removes pacman hooks before deleting packages so dkms remove is not triggered.


alkazar:
Nice!

PR #400, opened 2022-08-24
Remove openbox, using gnome now

Samsagax:


alkazar:
Still needed for the "switch window mode" chimera action.

Samsagax:
> Still needed for the "switch window mode" chimera action.

Mhhh... we should change that to use mutter instead I think, or just switch to GNOME.

alkazar:
The feature is useful as is for BPM. Once BPM is out of the picture this can be removed.
Alternatively, we could just drop this feature and tell people: go use the desktop.
The only problem is it may be difficult for some users to do that when the option to switch to desktop does not work in BPM.

Samsagax:
What about adding two shortcuts to steam: one for entering Desktop and one for switching to Gamescope/GampeadUI?

alkazar:
> What about adding two shortcuts to steam: one for entering Desktop and one for switching to Gamescope/GampeadUI?

Not sure I want to bother. Once BPM is out of the picture there won't be any need for that.
Closing for now.

PR #401, opened 2022-08-24
Remove boost-libs, should be a dependency

Samsagax:


alkazar:
So... IIRC this was added for compatibility with a specific game or games. We can try to remove it and see what breaks :sweat_smile:

PR #402, opened 2022-08-28
Remove xf86-video-intel Xorg driver

Samsagax:
Mesa is phasing out Intel driver in favor of modesetting since v20.0. We should remove it. Using modesetting is compatible with Nvidia prime configurations and iris graphics hardware.


alkazar:
Do we need to do something to enable modesetting?

Samsagax:
> Do we need to do something to enable modesetting?

No. It comes with Xorg and is the fallback driver for anything that it detects.

PR #404, opened 2022-09-01
Workaround: pygame rebuild

Samsagax:
This is a dirty hack to rebuild pygame and it is valid until ArchLinux maintainer rebuilds the package or the package gets updated (unknown time).


alkazar:
Don't think we need this anymore, I see there is an update to pygame.

PR #412, opened 2022-09-05
Use gamescope-plus-git

Samsagax:
This probably will fix the Mangohud not displaying in the correct position. Seems like the non-git version uses `wlroots` from the Arch repos while -git version uses a submodue with `wlroots` with a specific commit. It doesn't affect (visibly) on regular gamescope, but since we are rotating the screen on some handhelds it most certainly affect.


Samsagax:
Closing as this doesn't fix the issue as it did before

PR #425, opened 2022-10-19
Dockerfile: only use http or https

BoukeHaarsma23:
Since rsync is not installed within the docker-container reflector will spit errors at you. This forces reflector to use http or https.


Samsagax:
Shouldn't it be https for security?

BoukeHaarsma23:
It defaults to http,https and rsync. But I agree that https would make more sense

PR #426, opened 2022-10-20
Workaround: fatal: transport 'file' not allowed

BoukeHaarsma23:
On github and locally the python-leveldb package failed to build for me. Setting this worked around the issue for me.


Samsagax:
Relevant discussion in ArchLinux forum: https://bbs.archlinux.org/viewtopic.php?id=280571

PR #435, opened 2022-11-19
Freeze archive by date

Samsagax:
Freeze the repositories to the build date so when unlocking the image packages like the kernel won't break.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -222,6 +222,14 @@ echo "" >> ${BUILD_PATH}/build_info
 cat ${BUILD_PATH}/manifest >> ${BUILD_PATH}/build_info
 rm ${BUILD_PATH}/manifest
 
+# freeze archive date of build to avoid package drift on unlock
+export TODAY_DATE=$(date +%Y/%M/%d)
+echo "Server=https://archive.archlinux.org/repos/${TODAY_DATE}/\$repo/os/\$arch" > \
+${BUILD_PATH}/etc/pacman.d/mirrorlist
+
+cat ${BUILD_PATH}/etc/pacman.d/mirrorlist
+
+
 btrfs subvolume snapshot -r ${BUILD_PATH} ${SNAP_PATH}
 btrfs send -f ${SYSTEM_NAME}-${VERSION}.img ${SNAP_PATH}
 
--- a/build.sh
+++ b/build.sh
@@ -47,6 +47,13 @@ mkfs.btrfs -f ${BUILD_IMG}
 mount -t btrfs -o loop,nodatacow ${BUILD_IMG} ${MOUNT_PATH}
 btrfs subvolume create ${BUILD_PATH}
 
+# set archive date if specified
+if [ -n "${ARCHIVE_DATE}" ]; then
+	echo "
+	Server=https://archive.archlinux.org/repos/${ARCHIVE_DATE}/\$repo/os/\$arch
+	" > /etc/pacman.d/mirrorlist
+	pacman -Syyuu --noconfirm
+fi
 
 # build AUR packages to be installed later
 export GIT_ALLOW_PROTOCOL=file:https:git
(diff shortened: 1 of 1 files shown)


alkazar:
Is the goal to avoid kernel breakage? Or is there another purpose to this?

To avoid kernel breakage I have a better idea: on unlock we move the kernel files back to their default location and update the boot loader accordingly.

Samsagax:
> Is the goal to avoid kernel breakage? Or is there another purpose to this?
> 
> To avoid kernel breakage I have a better idea: on unlock we move the kernel files back to their default location and update the boot loader accordingly.

Not only kernel. But a lot of packages get borked and make the system unusable if for example glibc gets updated.
This way we can avoid a partial upgrade situation.
The usecase I'm envisioning is the user (or we ourselves) unlock a certain build for testing then isntall base-devel and git but the main distro drifted apart so much that a new version of glibc broke all other packages.

build.sh
```diff
@@ -222,6 +222,14 @@ echo "" >> ${BUILD_PATH}/build_info
 cat ${BUILD_PATH}/manifest >> ${BUILD_PATH}/build_info
 rm ${BUILD_PATH}/manifest
 
+# freeze archive date of build to avoid package drift on unlock
+# if no archive date is set
+if [ -z "${ARCHIVE_DATE}" ]; then
```
alkazar, line 227: My previous comments are not addressed.
 - ARCHIVE_DATE and TODAY_DATE could be different and cause problems.
 - The packages will be installed with the normal Arch Linux mirrors which may have different packages than TODAY_DATE.

Instead:
 - This condition should be the opposite. We should use the ARCHIVE_DATE only if set.
 - We need to make sure to set it before any packages are installed during the build.

Samsagax: I expect `ARCHIVE_DATE` and `TODAY_DATE` to be different. This is meant to not freeze the database to `TODAY_DATE` if `ARCHIVE_DATE` is already set and used (So it won't overwrite mirrorlist).

It is true that the archive could change a little from the mirrors and that would depend on the time of day they update both mirrors and the archive. I expect that change to be small enough to not cause major issues when unlocking the image.

This is only meant to be a sane default, the mirrorlist can still be modified by the user and use an up-to-date mirror. But if someone is good enough to know what they are doing, then he/she can handle the consequences.

Samsagax: You are not making people downgrade anything, the build.sh should be run inside a Docker image anyway. And if not, then it is at your own risk).
This works with two things:
- If no `ARCHIVE_DATE` is specified, then at the end of the build set the archive date to `TODAY_DATE` inside the image.
- if an `ARCHIVE_DATE` is specified, then downgrade the builder to match that and build all AUR packages against that.

The `pacstrap` step copies the mirrorlist from the builder to the image unless is explicitly requested.

alkazar: You misunderstand. I am worried about package downgrades (or maybe upgrades?) after doing a frzr-unlock because of the mismatch between the packages being installed from the main Arch repos and what is in the TODAY_DATE archive.

alkazar: Using TODAY_DATE would make more sense if it was an else on the ARCHIVE_DATE condition above.

Samsagax: > You misunderstand. I am worried about package downgrades (or maybe upgrades?) after doing a frzr-unlock because of the mismatch between the packages being installed from the main Arch repos and what is in the TODAY_DATE archive.

They should be upgrades and for whatever changed between our time of build and the end of the day when the archive is freezed for the next day. I expect to be minor changes, but we would never know. Still the package drift would be at most one day instead of weeks (or even months!). It is not a perfect solution in any way but is better than let package drift apart to break things.

We should be awarte that when we build and there is a huge package dump in the Arch repos afterwards in the same day, we should redo the image, but that is the current state anyway.

> Using TODAY_DATE would make more sense if it was an else on the ARCHIVE_DATE condition above.

It is an "else" implicitly, since that condition checks if `ARCHIVE_DATE` is set. If it isn't set, then go for `TODAY_DATE`. If it is set, then leave it alone and pass it, the builder would already have copied the mirrorlist to the image circa line 50. (Maybe I made a mistake here? I think the logic is correct)

PR #436, opened 2022-11-19
Dont uninstall on build

Samsagax:
Will add some speed to the build if we don't uninstall dependencies each time a package finishes building.


alkazar:
I fear this will overflow the limited space we have on the GitHub builder. We would have to at least be able to remove all the installed packages after all the AUR packages have been built.

Samsagax:
> I fear this will overflow the limited space we have on the GitHub builder. We would have to at least be able to remove all the installed packages after all the AUR packages have been built.

Never thought of that. Anyways the build packages are part of the cache we are currently copying over to the image system. I don't see how this will make the image grow.

PR #437, opened 2022-11-25
Add udev rule for ntfs3 driver

Samsagax:
This workarounds `mount` command without a explicit type declaration. Maybe the best solution comes in modifying the script to detect NTFS and mount it with explicit type, but this works it around and doesn't seem have any downsides so far.

Reference: https://wiki.archlinux.org/title/NTFS#unknown_filesystem_type_'ntfs'


alkazar:
Wouldn't this prevent other fs types from being mounted?

Samsagax:
> Wouldn't this prevent other fs types from being mounted?

Only affects ntfs partitions.

PR #454, opened 2022-12-14
Switch to using polkit for managing media in chimera app

pastaq:
steam-removable-media now supports formatting mmcblk0 using GamepadUI. Gives the user permission to run the script. The button is only active if an mmcblk device is detected by steam and can only format mmcblk0. The new script simply runs the old script with a hardcoded mmcblk0 parameter.


Samsagax:
LGTM.
But in the near future we should migrate all this sudoers fiddling into rootfs maybe?

pastaq:
> LGTM. But in the near future we should migrate all this sudoers fiddling into rootfs maybe?

I'd rather migrate to polkits tbh, they don't require username specification in the config as they are system wide policy. That would either enable us to have each package manage its own permissions, or we could install on rootfs directly.

pastaq:
> > I'd rather migrate to polkits tbh, they don't require username specification in the config as they are system wide policy. That would either enable us to have each package manage its own permissions, or we could install on rootfs directly.
> 
> YES! Could you do that here? I think the polkit for that script is already there.

Updated steam-removable-media to use polkits for the Format SD Card button. It will now work automatically on the next rebuild.

This changes the scope of this PR. It depends on https://github.com/ChimeraOS/chimera/pull/249 and is now to remove sudoers.d/chimera entry from build as it will not be necessary.

PR #456, opened 2022-12-16
Patch Mesa 22.3.1 to fix gamescope issues for Intel

ruineka:
We're using the Arch Linux's PKGBUILD including the additional ANV patch to fix the issues where gamescope would fail on Intel devices.

The patches only effect Intel specific files called `ISL_FORMAT` and `ANV_FORMATS`

### DIFF
--- a/pkgs/mesa-22.3.1-anv-patch/PKGBUILD
+++ b/pkgs/mesa-22.3.1-anv-patch/PKGBUILD
@@ -3,11 +3,17 @@
 # Contributor: Jan de Groot <jgc@archlinux.org>
 # Contributor: Andreas Radke <andyrtr@archlinux.org>
 
-pkgbase=mesa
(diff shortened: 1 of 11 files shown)


Samsagax:
If you install this as local packages, you should also remove mesa and all packages made by these PKGBUILDS from the manifest.

pkgs/mesa-22.3.1-anv-patch/PKGBUILD
```diff
@@ -71,7 +77,7 @@ build() {
     -D valgrind=enabled
 
   # Print config
- meson configure --no-pager build
+  meson configure --no-pager build
```
pastaq, line 80: Doh

Samsagax: > Doh

🤦🏻

PR #479, opened 2023-01-08
Add iomem=relaxed

pastaq:
Improves compatibility with ryzenadj for 6800u devices.

### DIFF
--- a/rootfs/usr/lib/frzr.d/bootconfig.conf
+++ b/rootfs/usr/lib/frzr.d/bootconfig.conf
@@ -1,1 +1,1 @@
-ibt=off split_lock_detect=off
+ibt=off split_lock_detect=off iomem=relaxed


Samsagax:
Is it advisable? Doesn't this introduce potential security issues and maybe EAC breakage?

pastaq:
I wouldn't run it on a server with a public facing IP, but it should be fine for our purposes. CONFIG_IO_STRICT_DEVMEM ensures if a kernel driver has reserved a device memory region for use, it will become unavailable to /dev/mem. The reservation in the kernel was to keep other kernel things from using the memory, but we want to do that with ryzenadj since that is essentially how it works. A malicious user or program would still need root access to /dev/mem to do anything. Ryzenadj-controller uses a Unix socket to give user access to this and sanitizes commands to ensure they are valid ryzenadj commands prior to execution, which is a reasonable concession IMO. 

I haven't been able to find any instances of anti-cheat breaking when I searched for it.

pastaq:
One other thing, the goal of this would be to prevent users from needing to manually change syslinux.cfg to get TDP control working properly on 6800u. Technically the settings will take without it, but the current values table cannot be dumped, so there's no way to validate a userspace tool is reflecting the actual TDP.

PR #486, opened 2023-01-11
Add goodix-gpdwin3 dkms driver

pastaq:
Add fixed goodix-gpdwin3-dkms driver with modified blacklist config. Can be enabled with device-quirks. Must install goodixgpdwin3.conf in /etc so device-quirks can edit it without frzr-unlock. This file will blacklist this driver so that it doesn't conflict with goodix_ts. Device-quirks will need to change this blacklist to goodix_tx. This will enable touch support for win3 devices.


alkazar:
Not sure I like having a whole driver in here. Can we not make this a separate package in the AUR?

pastaq:
> Not sure I like having a whole driver in here. Can we not make this a separate package in the AUR? Or at least obtain the driver source code from somewhere else?

There is a package in the AUR already, buy it won't work for our purposes. I suppose I could make this a git repo and the PKGBUILD here could build from there. Thoughts?

pastaq:
> Not sure I like having a whole driver in here. Can we not make this a separate package in the AUR? Or at least obtain the driver source code from somewhere else?

Updated. PKGBUILD now uses https://github.com/pastaq/goodix-gpdwin3-dkms as source and is tagged to the current HEAD.

Samsagax:
There are still the sources in this PR. Also seems to be a typo in the directory name?

pastaq:
> There are still the sources in this PR. Also seems to be a typo in the directory name?

Woops, forgot to push lol.

pastaq:
> There are still the sources in this PR. Also seems to be a typo in the directory name?

Fixed

PR #490, opened 2023-01-17
Remove r8168 driver

Samsagax:
Causes regressions with r8169 driver. The package blacklists r8169 kernel module and can't be reenabled automatically. Maybe if there are enough users with this chip, we should redo the packaging without the blacklisting part.

### DIFF
--- a/manifest
+++ b/manifest
@@ -145,7 +145,6 @@ export AUR_PACKAGES="\
 	python-inotify-simple \
 	python-vdf \
 	r8152-dkms \
-	r8168-dkms \
 	retroarch-autoconfig-udev-git \
 	rtl8812au-dkms-git \
 	rtl8814au-dkms-git \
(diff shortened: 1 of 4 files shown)


pastaq:
I think this can be managed with device quirks similar to the goodix-gpd-win3 driver. If we repackage to blacklist this by default and the autoconfigure script could detect hardware that needs this to swap the blacklist & modprobe. Is there a comprehensive list of hardware that needs the *68?

Samsagax:
We could do the same as with goodix but inverse. Just repackaging the same one and removing the modprobe blacklist file to be a user option under `/etc`.

> Is there a comprehensive list of hardware that needs the *68?

Not that I'm aware of.

Samsagax:
@pastaq this should allow for control the blacklist from `/etc` and still build the driver.

PR #492, opened 2023-01-20
Build packages on github

BoukeHaarsma23:
I have been using some version of this on my local builds for some time. Maybe it is a nice step towards splitting off the packages to a different submodule.

Let me know what you guys think.

### DIFF
--- a/manifest
+++ b/manifest
@@ -147,6 +147,7 @@ export AUR_PACKAGES="\
 	python-bmi160-i2c \
 	python-inotify-simple \
 	python-leveldb \
+	python-smbus2 \
 	python-vdf \
 	r8152-dkms \
 	retroarch-autoconfig-udev-git \


alkazar:
This is nice! Could it be extended to build AUR packages as well?

In any case, I do have a concern. Sometimes we freeze the packages by downloading from the Arch package archive. In that case, this could fail because the build environment of the packages would now be different than the installation environment.

Samsagax:
I don't get how this will improve builds? Does it caches the packages?

Samsagax:
Sounds cool. I'll give it a shot on my branch:)

Samsagax:
> Build (cached) now done in 37 minutes vs 1u 22minutes on latest successful build on master.

Neat!
Is it stable to sudden changes in those 37 minutes? I mean, the packages are all downloaded at once and built in the same replicated environment right?

Samsagax:
Might be worth to add a step betweent the override packages (aka, own packages) to install them on the builder before building all other AUR packages. That way the AUR build environment will closer to the installed image.

PR #496, opened 2023-01-29
Parallel pkg builds

BoukeHaarsma23:
I have implemented a docker image where all packages from the AUR are build in parallel.

I only tested it on github. So no 1 script to run everything yet.

Let me know what you guys think


freeze.sh
```diff
@@ -0,0 +1,51 @@
...
+# set archive date if specified and force update/downgrade builder
+if [ -n "${ARCHIVE_DATE}" ]; then
+	echo "
+	Server=https://archive.archlinux.org/repos/${ARCHIVE_DATE}/\$repo/os/\$arch
+	" > /etc/pacman.d/mirrorlist
+fi
+
+# Update system and install pikaur and other build dependencies
+echo -e "[multilib]\nInclude = /etc/pacman.d/mirrorlist\n" >> /etc/pacman.conf && \
+	pacman --noconfirm -Syyu && \
```
Samsagax, line 19: This needs another `-u` to force downgrade of installed packages after setting `ARCHIVE_DATE` to avoid package drifting.

freeze.sh
```diff
@@ -0,0 +1,51 @@
...
+		python-build \
+		python-installer \
+		python-setuptools \
+		&& \
+	pacman --noconfirm -S --needed git && \
+	echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
+	useradd build -G wheel -m && \
+	su - build -c "git clone https://aur.archlinux.org/pikaur.git /tmp/pikaur" && \
+	su - build -c "cd /tmp/pikaur && makepkg -f" && \
+	pacman --noconfirm -U /tmp/pikaur/pikaur-*.pkg.tar.zst
```
Samsagax, line 38: Any specific reason to do it in a multiline one-liner? Since the script is run with `set -e` any command that fails will have the same effect.

freeze.sh
```diff
@@ -0,0 +1,51 @@
...
+	chmod +x /usr/bin/systemd-run
+
+# Only run reflector when not using archive packages
+if [ -z "${ARCHIVE_DATE}" ]; then
+    pacman -S --noconfirm reflector
+    reflector --verbose --latest 20 --protocol https --sort rate --save /etc/pacman.d/mirrorlist
+fi
+
+# For packages that use git file protocol in docker container
+export GIT_ALLOW_PROTOCOL=file:https:git
```
Samsagax, line 51: This one shouldn't be needed anymore since all `-git` AUR packages are fixed.

manifest
```diff
@@ -101,6 +101,7 @@ export PACKAGES="\
 	podman \
 	pulsemixer \
 	python \
+	python-inotify-simple \
```
Samsagax, line 104: This should be in a separate PR since affects the distribution. Isn't this package a dependency already?

manifest
```diff
@@ -142,7 +143,6 @@ export AUR_PACKAGES="\
 	libretro-virtualjaguar-git \
 	nintendo-udev \
 	oxp-platform-dkms-git \
-	python-inotify-simple \
```
Samsagax, line 145: Same as above. Should be a separate PR.

freeze.sh
```diff
@@ -0,0 +1,51 @@
...
+
+set -e
+set -x
+
+source manifest
+
+# Allow multiple downloads
+sed -i '/ParallelDownloads/s/^#//g' /etc/pacman.conf
+# set archive date if specified and force update/downgrade builder
+if [ -n "${ARCHIVE_DATE}" ]; then
```
Samsagax, line 11: Maybe the freeze by `ARCHIVE_DATE` and the installation steps should be a different executable.
How the final archive freeze to date is handled in this scheme?

Samsagax (changes requested):
Looks good except for a few changes here and there. Good work!

.github/workflows/main.yml
```diff
@@ -4,7 +4,65 @@ on:
...
+    name: Build environment
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: docker/setup-buildx-action@v2
+      - name: build builder image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          tags: image-builder:latest
```
Samsagax, line 27: Maybe add the commit hash or the chimeraos version? This could be very useful for reproducible builds.

freeze.sh
```diff
@@ -0,0 +1,51 @@
...
+		xcb-util-wm \
+		fmt \
+		python-markdown-it-py \
+		python-wheel \
+		python-build \
+		python-installer \
+		python-setuptools \
+		&& \
+	pacman --noconfirm -S --needed git && \
+	echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
```
Samsagax, line 34: Is this neccesary? you are running inside docker as root. The only part that needs root is pikaur after build.

Samsagax:
@BoukeHaarsma23 I would really want to merge this one. Could you modify the things I mentioned?

PR #527, opened 2023-03-01
Bootloader migration away from syslinux

Samsagax:
This will need https://github.com/ChimeraOS/frzr/pull/32


rootfs/usr/lib/frzr.d/bootloader.migration
```diff
@@ -0,0 +1,23 @@
...
+
+  if ![ -d "$mount_path"/boot/syslinux.cfg ] ; then
+    return 0
+  fi
+
+  echo "Running post install bootloader migration"
+
+  if [ -d /sys/firmware/efi/efivars ]; then
+    echo "UEFI boot detected, installing systemd-boot"
+    bootctl install
```
ruineka, line 16: After this is ran we'll need to configure the loader and entries at `/EFI/loader/loader.conf` and `/EFI/loader/entries/chimeraos-${version}.conf`.

Here is a working example for an entry:
```
title   ChimeraOS v40
linux   /chimeraos-40_1238104/vmlinuz-linux
initrd  /chimeraos-40_1238104/amd-ucode.img
initrd  /chimeraos-40_1238104/initramfs-linux.img
options root=LABEL=frzr_root nvme_load=YES rw rootflags=subvol=deployments/chimeraos-40_1238104 quiet splash loglevel=3 rd.systemd.show_status=auto rd.udev.log_priority=3
```

alkazar:
`bootctl install` appears to change the UEFI boot OS list in the firmware. I dislike this because it makes it harder to move your installations around, or create a portable boot disk.

alkazar:
Nevermind, conclusion of offline discussion is let's go with using `bootctl install` for now. It should be more compatible and the issues it might cause are minimal.

joaquinvacas:
Also, AFAIK bootctl can sign the EFI loader for secure boot.

PR #547, opened 2023-03-29
Fix position issues with mangoapp

ruineka:
Confirmed to fix issues on devices that don't rely on specifying a `-w` `-h` nested value to work such as the intel One Xplayer 1S.

- Fixes the position of the performance overlay on the Aokzoe and Aya Neo Air


pastaq:
Tested on AYANEO AIR as working. Only fails on OneXPlayer 1S due to it not supporting its native resolution.

ruineka:
The latest commit seems to resolve the position of the performance overlay regardless of if `-w` and `-h` are used or not.

PR #550, opened 2023-03-30
Switch to polkit action from sudoers for brightness control

pastaq:


rootfs/usr/share/polkit-1/actions/org.chimeraos.priv.write.policy
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE policyconfig PUBLIC
+ "-//freedesktop//DTD PolicyKit Policy Configuration 1.0//EN"
+ "http://www.freedesktop.org/standards/PolicyKit/1/policyconfig.dtd">
+<policyconfig>
+
+  <vendor>ChimeraOS Chimera</vendor>
+  <vendor_url>http://chimeraos.org</vendor_url>
+
+  <action id="org.chimeraos.policykit.steamos.pkexec.run-set-hostname">
```
ShadowApex, line 10: Seems like this doesn't have a closing `</action>` tag. Should this line be removed?

PR #560, opened 2023-04-04
fix commitwrite/declinewrite

pastaq:
Previous version always "fails" to success. It has caused issues on users with AMD cards getting set to 15w by steam.

### DIFF
--- a/rootfs/usr/bin/steamos-polkit-helpers/steamos-priv-write
+++ b/rootfs/usr/bin/steamos-polkit-helpers/steamos-priv-write
@@ -13,15 +13,14 @@ WRITE_VALUE="$2"
 function CommitWrite()
 {
     echo "commit: $WRITE_VALUE -> $WRITE_PATH" | systemd-cat -t p-steamos-priv-write -p warning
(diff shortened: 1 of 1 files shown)


rootfs/usr/bin/steamos-polkit-helpers/steamos-priv-write
```diff
@@ -13,15 +13,14 @@ WRITE_VALUE="$2"
 function CommitWrite()
 {
     echo "commit: $WRITE_VALUE -> $WRITE_PATH" | systemd-cat -t p-steamos-priv-write -p warning
-    echo "$WRITE_VALUE" > "$WRITE_PATH"
     chmod a+w "$WRITE_PATH"
```
alkazar, line 16: what if $WRITE_PATH doesn't exist? `set -e` means the script would error out at this point.

pastaq: Write path should be a sysfs path. CommitWrite is only called if if matches a case from below in the file. I would expect it to fail if the path is missing.

PR #562, opened 2023-04-04
Mangohud update

BoukeHaarsma23:
See also https://github.com/flightlessmango/MangoHud/commit/bc282cf300ed5b6831177cf3e6753bc20f48e942

I have updated the PKGBUILD according to master, and updated the commit. It should now build without pinning the package. 
I also had to reformat the resolution.patch a bit.

### DIFF
--- a/manifest
+++ b/manifest
@@ -179,7 +179,6 @@ export AUR_PACKAGES="\
 "
 
 export PACKAGE_OVERRIDES="\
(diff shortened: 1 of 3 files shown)


alkazar:
Amazing. Thank you!

But can we keep the changes to `PACKAGE_OVERRIDES` and just remove the variable definition in manifest?
Next time we need to do something like this it would be good to have the fix in place.

PR #575, opened 2023-04-11
Add yay helper to make it easier to help users

ruineka:
It's become increasingly more difficult to try and help the users when issues happen and having an AUR helper can make it tremendously easier for us to help the users as well as them help us test changes to confirm fixes.

### DIFF
--- a/manifest
+++ b/manifest
@@ -177,6 +177,7 @@ export AUR_PACKAGES="\
 	xone-dkms \
 	xpad-noone-dkms \
 	xpadneo-dkms \
+	yay \
 "
 
 export SERVICES="\


npaladin2000:
yay is a little more widely used

ruineka:
> Is there a reason you chose yay over pikaur (or even yay-bin)?

I personally prefer pikaur, but it seems yay is the more widely used tool that people understand more.

alkazar:
I think I would prefer pikaur.

This would only be used by us, advanced users, or users who we give direction to, may as well be a tool that we are familiar with and prefer.

pikaur is also much faster to build (just python) and much smaller.

ruineka:
> I think I would prefer pikaur.
> 
> This would only be used by us, advanced users, or users who we give direction to, may as well be a tool that we are familiar with and prefer.
> 
> pikaur is also much faster to build (just python) and much smaller.

Done and done.

PR #577, opened 2023-04-11
Add caching to build system

BoukeHaarsma23:
1. Add an extra workflow that creates a new builder image daily (regardless of pkgbuild changes). This way we can follow arch-linux package updates. (AUR will always be built). This will run 2 am daily, or via manual trigger to fill a cache.
2. Implement caching based on PKGBUILD hash, if no changes are found the build will continue using the image built before. Skipping the endless mesa rebuild.
3. Implement PR checks to master, so building an image without releasing it.
4. Fix package overrides (added an if statement here).

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -1,0 +1,28 @@
+name: Build base environment
+on:
+  schedule:
+    - cron: "0 2 * * *"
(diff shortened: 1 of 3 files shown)


pastaq:
2am in what timezone?

PR #579, opened 2023-04-12
added tauri apps prerequisites

tsimouris:


tsimouris:
There should have been some more packages included, my bad. As per the reasoning, I've been working on these two projects:

- https://github.com/tsimouris/youtube-tv
- https://github.com/tsimouris/twitch-tv

Just need to configure the icons now and work on the tauri bundrer, for flatpak support that is. For now though AppImages can be used.

alkazar:
If the flatpak issue is resolved then we won't these packages?

tsimouris:
True, the solution is rather temporary till the flatpak bundler is ready; at that point these should be part of the flatpak dependencies

alkazar:
Looks like AppImage might also be a solution? In any case, I don't think we need native support for this. Most people will want to use Flatpak, and that would be the preferred solution.

PR #582, opened 2023-04-14
Upgrade mangohud to 0.6.9

oliverchang:
This is mainly to get
https://github.com/flightlessmango/MangoHud/commit/a155c10df39ee6d007779ed2ea65d191ff6aa3d7 which fixes a bug with the CPU wattage being shown as 1000x higher than what it actually is on my 6800u.


Samsagax:
Already done

PR #585, opened 2023-04-14
Strip mesa

BoukeHaarsma23:
Strips out old drivers (like r300, radeon etc), keeping the vulkan ones.
Strips out openCL drivers (do any games use this)

Removes offscreen and software rendering (this also requires to disable gallium-nine, but DXVK is used anyway right?)

@ruineka let me know what you think about this. Saves around ~500 compile passes.


ruineka:
Some games appear to use opencl such as BeamNG.drive.

ruineka:
https://en.m.wikipedia.org/wiki/List_of_OpenCL_applications#Games

alkazar:
This is interesting, but we should endeavor to reduce the amount of patches we have to maintain.
Since there is no benefit to the end-user, I would vote against merging this.

PR #588, opened 2023-04-15
local builds

BoukeHaarsma23:
This edits some scripts so it can skip tarring on local builds.

Also adds keeping the .cache folder so AUR builds will go a lot faster since it does not need to fetch sources all the time ;)


ruineka:
This is amazing. Thanks!

PR #591, opened 2023-04-16
only build intel vulkan driver

BoukeHaarsma23:
I came across the fact that [valve also uses a seperate radv driver](https://gitlab.com/evlaV/jupiter-PKGBUILD/-/blob/master/mesa-radv/PKGBUILD)

I thought why could we not do the same with our patched intel driver? So I tried this. No idea if it works, so looking for testers :)

### DIFF
--- a/manifest
+++ b/manifest
@@ -64,12 +64,15 @@ export PACKAGES="\
 	lib32-libxcrypt-compat \
 	lib32-libva \
 	lib32-libva-intel-driver \
+	lib32-libva-mesa-driver \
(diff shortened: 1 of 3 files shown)


ruineka:
The manifest would need to be updated to add all the other mesa packages that are being removed. This can be used as a reference point. https://github.com/ChimeraOS/chimeraos/commit/be81aab24cca0a923b6ae5cae12639d3a3d81f66

alkazar:
Neat!

PR #592, opened 2023-04-16
Update manifest - add rclone

npaladin2000:
add rclone package for future cloud sync support


alkazar:
I don't want to include packages for "future" things. Any new package should be directly connected to a feature users can actually make use of.

If you want to experiment, please use `sudo frzr-unlock` and install manually.

PR #637, opened 2023-05-26
Add wget for cryoutilities compatibility

ruineka:
Cryouilities makes use of wget during the install process.


alkazar:
Does it even work if installed though? Do we want users to be using it?

ruineka:
> Does it even work if installed though? Do we want users to be using it?

It did work in v40, it seems to no longer work. It's a very useful utility for those that need it or want it (seems to be a lot because of the Steam Deck). If we can figure out how to get it working I think that would be a good thing.

PR #652, opened 2023-06-26
use archive of yesterday to have some stable target

BoukeHaarsma23:
Had an idea to use the archive of yesterday.

This way, we can always sync up all the stuff before building anything.

Let me know what you guys think.


Samsagax:
> Seems like a good idea, thanks. My only concern would be slower build speed due to not using ranked mirrors.

That is true. But currently the download step is not that long. Maybe we could benchmark it?

pastaq:
What would this look like when pushing changes to our various packages during the 'bug fixes" part of testing? I think we'll need to formalize our process a little bit so packages don't get missed. We're already experiencing this some and I imagine this will compound the issues.

PR #667, opened 2023-07-09
Use submodules

BoukeHaarsma23:
Build is still flakey though, so please don't merge.

Started adding all AUR packages as submodules and using archive date of yesterday.

For testing we could then pin Arch packages and start adding/removing packages.

### DIFF
--- a/build-image.sh
+++ b/build-image.sh
@@ -82,9 +82,11 @@ pacman-key --populate
 echo "LANG=en_US.UTF-8" > /etc/locale.conf
 locale-gen
 
+# Disable parallel downloads
+sed -i '/ParallelDownloads/s/^/#/g' /etc/pacman.conf
 
 # update package databases
-pacman --noconfirm -Syy
+pacman --noconfirm -Syyuu
 
 # install kernel package
 if [ "$KERNEL_PACKAGE_ORIGIN" == "local" ] ; then


manifest
```diff
@@ -5,7 +5,7 @@ export SYSTEM_DESC="ChimeraOS"
 export SYSTEM_NAME="chimeraos"
 export USERNAME="gamer"
 export SIZE="10000MB"
-export ARCHIVE_DATE=""
+export ARCHIVE_DATE=$(date -d 'yesterday' +%Y/%m/%d)
```
Samsagax, line 8: I guess this will be replaced by a concrete date down the line.

.github/workflows/build.yml
Samsagax, line 1: I like to keep this file reusable. The thing we would need is to pass on the mirrorlist with the archive date set and force upgrade (really downgrade) all packages before pushing.

Btw, do we need to upload to the registry? Can't we use the artifact uploaded to the action outputs? Seems more compact and less cumbersome. But if it works as is, then keep it.

.github/workflows/main.yml
Samsagax, line 1: I think you wanted to integrate the builder step here but taht makes the other changes less obvious (or I'm the kind of idiot that can't tell them apart, it happens). Could you undo the builder merge or make it in a separate commit?

manifest
```diff
@@ -180,6 +180,7 @@ export AUR_PACKAGES="\
 	libretro-prosystem-git \
 	libretro-stella2014-git \
 	libretro-virtualjaguar-git \
+	kanit-font \
```
Samsagax, line 183: Is this change needed here? If the package name changed, please put it in a separate commit (and push it to master!)

pkgs/build-package.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -x
+
+sudo chown -R build:build /workdir/pkgs
+
+PIKAUR_CMD="PKGDEST=/workdir/pkgs pikaur --noconfirm -S -P /workdir/${1}/PKGBUILD"
+PIKAUR_RUN=(bash -c "${PIKAUR_CMD}")
+"${PIKAUR_RUN[@]}"
```
Samsagax, line 9: Might be worth migrating all to makepkg instead of relying on pikaur. Can be tedious but can be simply scripted and reused if done right.

pkgs/build-packages.sh
Samsagax, line 1: Seems like a simple rename. Is there a reason? Can you do this in a separate commit?

utils/refresh-aur-packages.sh
```diff
@@ -0,0 +1,14 @@
+# /bin/bash
+
+source manifest
+pushd "aur-pkgs/" > /dev/null
+# remove all packages
+git rm -- */ > /dev/null || git commit -m "Update AUR packages" > /dev/null
+
+for package in ${AUR_PACKAGES}; do
+    git submodule add --force https://aur.archlinux.org/${package}.git > /dev/null
```
Samsagax, line 9: You don't need to check them out, just declare them, maybe don't use `--force`?

Anyway, can you separete this script to a different commit? Seems a sef contained change.

Dockerfile
```diff
@@ -32,7 +32,8 @@ RUN echo -e "#!/bin/bash\nif [[ \"$1\" == \"--version\" ]]; then echo 'fake 244
 
 COPY manifest /manifest
 # Freeze packages and overwrite with overrides when needed
-RUN source /manifest; if [ -n "${ARCHIVE_DATE}" ]; then echo "Server=https://archive.archlinux.org/repos/${ARCHIVE_DATE}/\$repo/os/\$arch" > /etc/pacman.d/mirrorlist; else reflector --verbose --age 2 --latest 20 --protocol https --sort rate --save /etc/pacman.d/mirrorlist; fi && \
+RUN source /manifest && \
```
Samsagax, line 35: Why removing this? Makes the Dockerfile more versatile, making it work for both frozen and non-frozen archive date

manifest
Samsagax, line 1: This whole commit should be in master

build-image.sh
Samsagax, line 1: Shoud not do `-Sy` without `-u`. You could end up in a [partial upgrade situation](https://wiki.archlinux.org/title/system_maintenance#Partial_upgrades_are_unsupported)

Samsagax (changes requested):
I think this can and should be improved. Such a big change needs more care and eyes on them. I' ll take it as a proof of concept. If you beat me to the solution, please do(I have limited time to work on this until tonight).

In general the workflow should:

- Developer branches off from master were the stable release was made (e.g. `stable-43`)
- Convert all AUR packages to submodules to keep them in the current state and update them if needed for a specific fix.
- Set the ARCHIVE_DATE to the day were the release was built (not neccesarily the date it was promoted to stable)
- Builder image should make each package as a submodule and then redo the same image.

That last step I don't see clearly how to do it, maybe check if we are in master and reuse the same submodule idea and update all packages that way we use the same pipeline for building.

PR #670, opened 2023-07-12
Adjust session returning 

pastaq:
There is an issue with having all the "switch back to" files at the same time. Primarily, each session has a different requirement for what services should exists and the state of the filesystem. Miss-alignment of these has led to users unknowingly selecting the wrong session and reporting hardware failures when the issue is that the services for those sessions aren't running. Fixing this requires amending the chimeraos rootfs, gamescope-session, and opengamepadui-session to align with a common pattern.

- Adjust chimera-session to link selected sessions .desktop file to .local/share/applications
- Add OpenGamepadUI session switcher .Desktop files

~~-Move gamescope-session specific files to gamescope session~~

1.) Selecting any session with chimera-session will clear session-return.desktop from `~/.local/share/applications`
2.) Selecting gamesope-session, opengameadui-session, or qamepadui-with-qam-session will link the appropriate .desktop file to `~/.local/share/applications/session-return.desktop`

This will prevent a user from leaving the gamepadui-with-qam session and entering the gamescope session without HandyGCCS running and all the input devices hidden, which is the most common issue currently.

~~Depends on the following PR's being merged:~~
~~https://github.com/ChimeraOS/gamescope-session/pull/52~~
~~https://github.com/ShadowBlip/OpenGamepadUI-session/pull/4~~


pastaq:
Adjusted this PR. After thinking about it, it didn;t make sense to have chimeraos specific .Desktop files in the gamescope-session and opengamepadui-session packages. Moved them here into a folder `/usr/share/return-sessions/`

build-image.sh
```diff
@@ -203,6 +203,10 @@ rm -rf ${FILES_TO_DELETE}
 mkdir /home
 mkdir /var
 mkdir /frzr_root
+
+# Add return-session link for default session
+mkdir -p /home/${USERNAME}/.local/share/applications/
+ln -s /usr/share/return-sessions/org.chimeraos.Gamescope.Desktop /home/${USERNAME}/.local/share/applications/session-return.Desktop
```
Samsagax, line 209: `session-return.Desktop` -> `session-return.desktop` (should be lowercase `d`)

manifest
```diff
@@ -194,7 +194,6 @@ export AUR_PACKAGES="\
 	rtl8821au-dkms-git \
 	rtw89-dkms-git \
 	ryzenadj-git \
-	ryzenadj-controller-git \
```
Samsagax, line 197: Shouldn't this be part of its own patch?

manifest
```diff
@@ -214,15 +213,13 @@ export SERVICES="\
 	handycon \
 	haveged \
 	lightdm \
-	ryzenadj-controller \
```
Samsagax, line 217: Shouldn't this be on its own patch?

manifest
```diff
@@ -214,15 +213,13 @@ export SERVICES="\
...
 	sshd \
 	systemd-timesyncd \
 "
 
 export USER_SERVICES="\
 	chimera.service \
 	gamemoded.service \
-	ryzen-tctl \
```
Samsagax, line 225: Same with this one. Shouldn't it be its own patch?

rootfs/usr/bin/chimera-session
```diff
@@ -66,25 +69,28 @@ function desktop_xorg() {
 function gamepadui() {
 	# switch to Gamepad/Deck UI (Gamescope/Wayland)
 	echo '' > ${SESSION_CONFIG}
+	clean_steam_shortcuts
+	ln -s /usr/share/return-sessions/org.chimeraos.Gamescope.desktop ${DATA_HOME}/applications/session-return.desktop
```
Samsagax, line 73: `rg.chimeraos.Gamescope.desktop` < should be lowercase G?

rootfs/usr/bin/chimera-session
```diff
@@ -66,25 +69,28 @@ function desktop_xorg() {
...
 	sudo chimera-session-use-lightdm gamescope-session
 	sudo systemctl enable --now handycon
-	sudo systemctl enable --now ryzenadj-controller
```
Samsagax, line 71: Why removing ryzenadj-controller?

Samsagax (commented):
I'm not entirely sure this is the best approach. Maybe is better to write something to the sentinel file (currently we are not writing anything but is read) and use the same gnome-session-oneshot script to select to which session to return.

rootfs/usr/bin/chimera-session
pastaq: Its broke.

rootfs/usr/bin/chimera-session
pastaq: I left the case as it was, but we can change it

build-image.sh
pastaq: same as the other, retained historical case

pastaq:
> I'm not entirely sure this is the best approach. Maybe is better to write something to the sentinel file (currently we are not writing anything but is read) and use the same gnome-session-oneshot script to select to which session to return.

I'm not sure what that would look like tbh, I have no experience with it.

Samsagax:
> > I'm not entirely sure this is the best approach. Maybe is better to write something to the sentinel file (currently we are not writing anything but is read) and use the same gnome-session-oneshot script to select to which session to return.
> 
> I'm not sure what that would look like tbh, I have no experience with it.

I dontk know if it's even possible. I'll look into it but if this works let's use it as a workaround for the moment.

PR #673, opened 2023-07-15
Reproducible builds part 2

BoukeHaarsma23:
This add the part with checking out the submodules. The release flow would be something like:
- We test using unstable untill we want to freeze the packages and start testing cycles
- Run the release workflow from github web. This creates the stable-XX branch where all images get pushed to testing immediately. This will also setup all AUR packages as submodules at the point of branching, and fixates the archive_date. It will also bump version on master (without triggering a build)
- The first build (on stable-XX) is not easily triggerable from the previous step without some real hacky work-arounds with sharing PAT tokens or making the workflow really complex. So I opted for a manual trigger here from web interface as wel (or an empty commit on the stable-xx branch would also work off course). Both release/master branch share the image generation part of the system build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
 
 jobs:
   build-docker-image:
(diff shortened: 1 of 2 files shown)


Samsagax:
LGTM!

If there are no objections I'll merge this tomorrow

Samsagax (approved):

PR #675, opened 2023-07-16
Use a custom makepkg.conf for building and include it in the final image

NeroReflex:
This patch include the makepkg.conf file, used when compiling AUR software in the repository, providing the ability to change how the software is compiled and that same file is included in the resulting image.

For the project this has the benefit of reducing overall build time as in the options check is excluded, therefore resulting executables are not tested.

The rest of the file is equal to the default one I found in the docker container it was used to build chimera.

The main reason I did this is to prevent executing AUR software in the container as that software might be incompatible with the build environment.


Dockerfile
```diff
@@ -30,6 +30,7 @@ RUN echo -e "keyserver-options auto-key-retrieve" >> /etc/pacman.d/gnupg/gpg.con
 RUN echo -e "#!/bin/bash\nif [[ \"$1\" == \"--version\" ]]; then echo 'fake 244 version'; fi\nmkdir -p /var/cache/pikaur\n" >> /usr/bin/systemd-run && \
     chmod +x /usr/bin/systemd-run
 
+COPY rootfs/etc/makepkg.conf /etc/makepkg.conf
```
Samsagax, line 33: What is exactly changed from the default? Can it be just a `sed` over the specific line on `makepkg.conf`? My concern is upstream changes can go unnoticed in other areas if we freeze the file in our repo. Might be better to search/replace that specific line.

rootfs/etc/makepkg.conf
```diff
@@ -0,0 +1,159 @@
...
+# Makepkg defaults: BUILDENV=(!distcc !color !ccache check !sign)
+#  A negated environment option will do the opposite of the comments below.
+#
+#-- distcc:   Use the Distributed C/C++/ObjC compiler
+#-- color:    Colorize output messages
+#-- ccache:   Use ccache to cache compilation
+#-- check:    Run the check() function if present in the PKGBUILD
+#-- sign:     Generate PGP signature file
+#
+BUILDENV=(!distcc color !ccache check !sign)
```
Samsagax, line 68: Isn't it supposed to have `!check` here?

Samsagax (changes requested):

rootfs/etc/makepkg.conf
NeroReflex: Yes, good spot. I tested this in a fork that targets a zen4 cpu, built with zen3 cpu, so I know disabling check works, but I have forgotten to include that. Updated

Dockerfile
NeroReflex: I have now included a check that will fail the docker container construction if the upstream makepkg.conf gets changed so it won't go unnoticed.

As for why it is not a simple sed command, well this way the file is in the repo for reference and the process of changing something is straighforward. Moreover, speaking of upstream changes, with sed you will never know if you substituted more than what was intended to just because of upstream change that went unnoticed.

NeroReflex: The checksum check additional step refers to a /etc/makepkg.conf file that was calculated from a clean archlinux:base-devel docker container. Once "!check" is restored to "check" the hash will match proving That was the only modification that was done to the one you are currently using.

Samsagax (changes requested):

build-image.sh
```diff
@@ -47,6 +47,9 @@ mkfs.btrfs -f ${BUILD_IMG}
 mount -t btrfs -o loop,nodatacow ${BUILD_IMG} ${MOUNT_PATH}
 btrfs subvolume create ${BUILD_PATH}
 
+# copy the makepkg.conf into chroot
+cp /etc/makepkg.conf rootfs/etc/makepkg.conf
```
Samsagax, line 51: This is unnecesary. We wouldn't want to force users our configuration.

.gitignore
```diff
@@ -5,4 +5,5 @@
 output/*
 aur-pkgs/*
 !aur-pkgs/build-aur-packages.sh
-!aur-pkgs/build-aur-package.sh
\ No newline at end of file
+!aur-pkgs/build-aur-package.sh
+rootfs/etc/makepkg.conf
```
Samsagax, line 9: Don' t add unrelated changes. Split it up to other PR. Only the sed change should be made here

Samsagax (changes requested):

.gitignore
NeroReflex: This wasn't meant to end up in this merge request, this means something I did not intended ended up on this branch. I am revoking the merge request

PR #677, opened 2023-07-16
Add tools to update device firmware

majamee:
I noticed that in the current version of Chimera OS is no method present to update device firmware. Therefore I added "gnome-firmware" to the manifest of ChimeraOS. This tool should allow every user from the desktop session to easily and safely update their device firmware. If you do not wan't to include a GUI tool please at least include the underlying "fwupd" tool instad of "gnome-firmware".

### DIFF
--- a/manifest
+++ b/manifest
@@ -44,6 +44,7 @@ export PACKAGES="\
 	gnome-console \
 	gnome-control-center \
 	gnome-disk-utility \
+	gnome-firmware \
 	gnome-keyring \
 	gnome-session \
 	gnome-shell \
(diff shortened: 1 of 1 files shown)


alkazar:
If we include fwupd I think Gnome Software can handle firmware updates? If so, we wouldn't need to include a separate tool just for that.

alkazar:
Literally as I typed my previous comment, Gnome Software on my desktop popped up with a firmware update notification.

majamee:
Then it's perfect to just include fwupd. Shall I adjust the PR or do you want to?

majamee:
@alkazar done :)

Samsagax:
LGTM

PR #678, opened 2023-07-16
Remove ryzenadj-controller and don't enable ryzen-tctl

pastaq:
python-asyncio recently removed coroutine from the library. This breaks the message handling in ryzenadj-controller. Remove ryzenadj-controller for now until I can be bothered to fix it.


Samsagax:
What does this fix aside from ryzenadj-controller failing?

PR #683, opened 2023-07-18
Xpadneo fix

BoukeHaarsma23:
Have not been able to test this.


pkgs/xpadneo-dkms/drop-etc-files.patch
```diff
@@ -0,0 +1,38 @@
+diff -crB hid-xpadneo/dkms.post_install hid-xpadneo.new/dkms.post_install
+*** hid-xpadneo/dkms.post_install	2023-03-01 02:13:11.483605532 -0300
```
alkazar, line 2: what is the purpose of this patch?

PR #693, opened 2023-07-21
Fix session switcher return to the correct gamepad

pastaq:


pastaq:
The session-return desktop file still returns to gamescope-session every time.

rootfs/usr/bin/gnome-session-oneshot
```diff
@@ -5,12 +5,41 @@ set -e
 die() { echo >&2 "!! $*"; exit 1; }
 
 SENTINEL_FILE="steamos-session-select"
+RETURN_SESSION_FILE="chimeraos-session-return"
```
pastaq, line 8: What is this?

Samsagax: A file Introduced to keep the name of the session we should return to. But I most likely screwed up it's location on write

rootfs/usr/lib/os-session-select
```diff
@@ -39,9 +40,18 @@ if [[ -z $SENTINEL_CREATED ]]; then
...
+    mkdir -p "$config_dir"
+    cd "$config_dir"
+    echo "$return_session" > "$RETURN_SESSION_FILE"
+    cat "$RETURN_SESSION_FILE"
+  )
+
   # clear steam game desktop shortcut clutter
   DATA_HOME=${XDG_DATA_HOME:-$HOME/.local/share}
-  grep --files-with-matches "Exec=steam steam://rungameid/" ${DATA_HOME}/applications/* | tr '\n' '\0' | xargs -0 -I {} rm {} || true
+  grep --files-with-matches "Exec=steam steam://rungameid/" "${DATA_HOME}"/applications/* | tr '\n' '\0' | xargs -0 -I {} rm {} || true
```
pastaq, line 54: We seem to do this twice. It's also in chimera-session. I added a bug fix to that line where I removed the * and added a -r to the beginning of the grep.

PR #712, opened 2023-07-30
Update mangohud

Samsagax:
Includes the patch for k10temp from @BoukeHaarsma23


Samsagax:
> Zenergy also got merged. Maybe better to drop the kernel patches and use [zenergy AUR package](https://aur.archlinux.org/packages/zenergy-dkms-git)

Is not a bad idea. I prefer a closed package, do one thing and one thing only, instead of patching a driver not designed for it.

PR #716, opened 2023-07-31
Pull Dave Reisner's key and use pacman's keyring

elril0000:
The chimera pkg builds successfully with these changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ LABEL contributor="shadowapex@gmail.com"
 COPY rootfs/etc/pacman.conf /etc/pacman.conf
 RUN echo -e "keyserver-options auto-key-retrieve" >> /etc/pacman.d/gnupg/gpg.conf && \
     pacman-key --init && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -3,6 +3,7 @@ LABEL contributor="shadowapex@gmail.com"
 COPY rootfs/etc/pacman.conf /etc/pacman.conf
 RUN echo -e "keyserver-options auto-key-retrieve" >> /etc/pacman.d/gnupg/gpg.conf && \
     pacman-key --init && \
+    pacman-key --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 1EB2638FF56C0C53 && \
```
Samsagax, line 6: Van this be avoided if `auto-key-retrieve` is in effect with the change below?

elril0000: I just tested it and I think so. I'll push the change

Samsagax: Great

Samsagax:
Thanks for your contribution

PR #717, opened 2023-08-02
Add default config for wireplumber

ruineka:
This seems to be beneficial on all devices I've messed with. This won't fix all hardware unfortunately, but it should be a much better default for most hardware configurations when gamescope is being used.


pastaq:
Do we need to remove the audio fix for device-quirks at the same time as this?

ruineka:
> Do we need to remove the audio fix for device-quirks at the same time as this?

No, because 96k is still required for some devices and not others.

PR #736, opened 2023-08-21
Gpu reset

BoukeHaarsma23:
Requires https://github.com/ChimeraOS/linux-chimeraos/actions/runs/5930538847


pastaq:
The udev rule would be better located in the device-quirks package IMO. We should also migrate the remaining udev rules.

PR #743, opened 2023-08-27
remove nvidia drivers

alkazar:


pastaq (approved):

flukejones:
Please don't remove nvidia drivers. It appears that the latest versions 535.104.05, Steam in big picture mode runs very well with 2 caveats:
- it runs super well if started in big picture mode, and seems to get progressively worse *only* if switching to desktop style UI and back to BP
- There is a minor glitch when showing left settings bar (desktop background shows briefly)

I know nvidia is a PITA because of certain things but they are definitely improving.

PR #746, opened 2023-08-29
Yafti

BoukeHaarsma23:
As a placeholder for now, while we think about what we want to add here.

Ideas:
- Emudeck
- Notify user when gamescope-session has gone wrong
- ...

Open for suggestions here.


pastaq:
I need some background on this. Is it an automated script running thing of some kind?

KyleGospo:
> Yafti stands for Yet Another First Time Installer. It is also what bazzite uses in their desktop mode.
> [Upstream url](https://github.com/ublue-os/yafti)
> 
> When starting desktop mode a window will pop up with yafti (only the first time, unless you want it to obvs). The idea is that yafti can help install some often used utilities (like cryotools and the like) in a one-click fashion.
> 
> Now that distrobox is added I'll see if I can rework this a bit.

Right now this works well as a first time installer, but if you want to use it as a configurator where it knows the state of things that were installed in the past and can uninstall or reinstall things, that's something that we want to work on for a future release. 

If you're open to collaborating on it we're more than happy to do that, just message us in either Chimera or Universal Blue discord.

dragoonDorise:
Hi, EmuDeck creator here, I would love to help on adding EmuDeck and even tailor the experience for ChimeraOS users

pastaq:
> Hi, EmuDeck creator here, I would love to help on adding EmuDeck and even tailor the experience for ChimeraOS users

That sounds great. It might be best to coordinate on Discord. Are you on the ChimeraOS Discord? I'll send you an invite to a private chat where we can discuss this in detail.

PR #751, opened 2023-09-03
Add steam_notif_daemon package

Samsagax:


pastaq:
What does it do?

Samsagax:
> What does it do?

Not sure what Valve is using it for. But it sends to the Steam client all notifications the dbus produces. I assume is for showing a toast when something happens in the system

pastaq:
Interesting. I was thinking they were directly reading the dbus messages like we do.

alkazar:
Do we have to also enable it?

Samsagax:
> Do we have to also enable the service?

I already added the needed line in gamescope-session in a generic way (with typos an all): https://github.com/ChimeraOS/gamescope-session/commit/a4a7275ff52c6448d888f7bc186768345d20f0fc

DocMAX:
i try to send with "notify-send" but nothing happens. How is this supposed to work?

Samsagax:
You don't really need this to send notifications to steam. You can use the http API like the daemon does:

https://github.com/Joshua-Ashton/simple_notif_daemon/blob/0ba7aef86eec3efda037258ccb8bdf1ae357a9b9/main.c#L91

PR #778, opened 2023-10-10
Improve terminal with bash-completion for standard autofill

### LINKED ISSUE #777: Feature request: Add "bash-completion" dependency
Hey guys, if you add `bash-completion` to the base image, users will finally have a useful terminal with Bash completion support. Tested, and no issues here.

TriVoxel:
This package massively improves the terminal. It allows for pressing tab to autofill commands, subcommands, and files, just like a normal distro.

Adds ~0.83 MiB

Closes #777

### DIFF
--- a/manifest
+++ b/manifest
@@ -18,6 +18,7 @@ export PACKAGES="\
 	alsa-firmware \
 	alsa-utils \
 	amd-ucode \
+	bash-completion \
 	broadcom-wl-dkms \
 	cifs-utils \
 	diffutils \


pastaq:
This is something that I would use regularly. 90% of my time is actually spent in the terminal so improved auto-completion would be a decent QOL improvement.

pastaq (approved):

TriVoxel:
> I understand it is useful for our personal use. But is it useful for our target audience? I really think we should not make the image too bloated just to cover everyone's usecase. I think we should keep focussed on 'just a gaming' os. I personally feel that this and all the other PR's add up and is a slippery slope.

Yes, this is good for the average user. Everyone will use the terminal at some point. Without this (very small) package, the terminal is very annoying to use. The features I've proposed to add (and remove) are the only ones I think it's really missing to add the final level of polish. A total of 11-14 MiB of additional space. Trust me, this will help the average user as when they press tab, it will autocomplete in useful ways that bash feels broken without.

TriVoxel:
I think most noobs will not use SSH. Bash completion can hardly be classified as bloat. This barely affects the scope of the project, only adds the last remaining polish needed for desktop users to be satisfied. That is my goal, in line with your own. I just want this to be an all-around good gaming experience for new Linux users, that includes a high quality, bare minimum desktop environment without huge missing features, bloat apps... just the stuff people need to use the desktop comfortably and to get a full experience. It isn't that hard, just need to spare 7 PRs and +11-14 MiB of disk space.

npaladin2000:
I have to agree, this isn't meant to be a desktop gaming OS, it's meant to be a console gaming OS. The desktop is not there to be used as a desktop, it's there because there are some things that just can't be done without a desktop, including changing several Steam settings and running EXE installers.  If people want a desktop OS to game on there are much better options out there.

pastaq:
This is less than a Meg and we regularly direct users to the terminal, including TTY so it's not desktop specific. I don't see an issue with this change.

alkazar:
If we can improve autocompletion, I think that is worth having. But what I don't understand is what this package actually does. Autocompletion already works for me in the terminal (I hit tab and things like file/folder names get autocompleted).

What specifically does this add? Any examples?

pastaq:
> If we can improve autocompletion, I think that is worth having. But what I don't understand is what this package actually does. Autocompletion already works for me in the terminal (I hit tab and things like file/folder names get autocompleted).
> 
> What specifically does this add? Any examples?

It doesn't enable tab completion as a concept, it adds tab completion to many existing programs. We could even extend this for chimera-session fairly easily.

https://github.com/scop/bash-completion/tree/master/completions

TriVoxel:
pastaq is precisely right. Think pacman, flatpak, distrobox, and many, many others. Many CLI applications take advantage of bash-completion to allow the user to press tab, and then the program intelligently auto-fills data. For instance, type `git push origin <partial branch name>` and hit tab. It will complete it. Same with typing `flatpak run` followed by tab, it will print a list of applications you can run, and will autofill when you start typing one out and press tab. That is very program specific. Basically, it allows for context-sensitive completion as defined by the CLI application's developer. Without it, you only get  rudimentary file path and bin completions. Basically, without this, think of every single CLI application's usability being kneecapped and missing completion features.

PR #781, opened 2023-10-10
Remove File Roller as it duplicates built-in Nautilus functionality (obscolete package)

TriVoxel:
This PR removes File roller as it is obscolete due to Nautilus natively having both compression and decompression features. Adds duplicate functionality. This cleans up the install a bit.

Also added `p7zip` and `unrar` packages. These serve to allow Nautilus to natively compress and decompress `.7z` and `.rar` files natively.

Frees >=4.11 MiB

Screenshot from 2023-10-09 18-34-32

Screenshot from 2023-10-09 18-35-46
Screenshot from 2023-10-09 18-37-28

### DIFF
--- a/manifest
+++ b/manifest
@@ -19,6 +19,7 @@ export PACKAGES="\
 	alsa-utils \
 	amd-ucode \
 	broadcom-wl-dkms \
+	bzip2 \
 	cifs-utils \
 	diffutils \
 	dkms \
@@ -30,6 +31,7 @@ export PACKAGES="\
 	ethtool \
 	evtest \
 	ffmpeg \
+	file \
 	firejail \
 	flatpak \
 	fmt \
@@ -47,6 +49,7 @@ export PACKAGES="\
 	gnome-software \
 	gnome-text-editor \
 	gst-plugin-pipewire \
+	gzip \
 	haveged \
 	htop \
 	intel-gpu-tools \
@@ -94,6 +97,7 @@ export PACKAGES="\
 	linux-firmware \
 	liquidctl \
 	logrotate \
+	lrzip \
 	lshw \
 	mesa-demos \
 	modemmanager \
@@ -121,7 +125,9 @@ export PACKAGES="\
 	sshfs \
 	steam \
 	sudo \
+	tar \
 	ttf-liberation \
+	unace \
 	unrar \
 	unzip \
 	usb_modeswitch \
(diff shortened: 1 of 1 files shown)


pastaq:
Why were 7zip and unrar added?

TriVoxel:
> Why were 7zip and unrar added?

Nautilus needs `p7zip` to extract and compress 7-Zip archives. Nautilus needs `unrar` to decompress .rar archives.

pastaq:
My hesitation on this is due to not knowing for sure what we're using fileroller for at this time. I thik this package predates us adding a desktop at all. It is possible that its a soft dependency for something we're doing with the chimera app or one of its functions like wyvern support or un-packaging ROM's. A lot of these packages were added a long time ago when our dependency tracking didn;t need to be as sophisticated as it does now. If this is called by name through a bash call in a script or python binding somewhere that could break, so we'll need some more eyes on this to make sure we don't cause bigger headaches down the road.

TriVoxel:
Sounds good. AFAIK, File Roller is just a GUI front end for packages `xz`, `gzip`, `tar`, `p7zip`, and `unrar`. I bet if we manually install those, we will have the full CLI archive management system that File Roller depends on. But yeah, better safe than sorry to be sure.

Adding the aforementioned packages should give us the full CLI archive suite though.

alkazar:
Archive manager was only included because it seemed like a useful basic tool. It still might be useful. I suppose it allows you to inspect archives without decompressing them which I don't think Nautilus can do. Also, it may not be clear to people that you must use Nautilus to extract archives.

TriVoxel:
Archive manager was only included because it seemed like a useful basic tool. It still might be useful. I suppose it allows you to inspect archives without decompressing them which I don't think Nautilus can do. Also, it may not be clear to people that you must use Nautilus to extract archives.

It is true that Nautilus can't preview archives. However, it is much simpler. Simply double-clicking the archive, pressing enter, or using the context menu allows for extraction. Right clicking 1 or more files allows for compressing to multiple formats via context menu.

Futhermore, File Roller can be installed from Flatpak for those who want to preview archives, though I find those situations exceedingly rare myself. I think most users only download archives with the intention of extracting them, and the archive still needs to be partially decompressed to preview it, too, taking about as long to preview as it takes to fully extract it in my experience. Users who want the expected Nautilus fast extraction feature will need to change the default program from File Roller for each extention.

pastaq:
I don't see a problem with this.

PR #782, opened 2023-10-11
Enable video thumbnail previews in file manager

TriVoxel:
This PR adds only 0.26 MiB of size. It allows the Nautilus file manager to generate video thumbnails. You could imagine this being nice especially for Discord users, for instance.

Before:

Screenshot from 2023-10-11 00-57-32

After:

Screenshot from 2023-10-11 00-56-44

Size:

image

### DIFF
--- a/manifest
+++ b/manifest
@@ -30,6 +30,7 @@ export PACKAGES="\
 	ethtool \
 	evtest \
 	ffmpeg \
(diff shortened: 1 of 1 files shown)


alkazar:
Thanks. I think this is a good one to have.

PR #821, opened 2023-11-27
Add efibootmgr to support dual booting with Windows

yisiqi:
Add efibootmgr to support dual booting with Windows


yisiqi:
By adding the efibootmgr, we can use cli or desktop shortcuts to restart the system and boot to Windows Boot Manager. This is very useful for dual system users.

> How does adding this help with dual booting Windows?

alkazar:
Got it. Thank you.

PR #829, opened 2023-12-12
Tailscale 

diericx:
Tailscale+Syncthing is a very common stack for syncing files between computers and it works very well for syncing saves between a Steam Deck and a Gaming PC over LAN or WAN. 

I had a lot of trouble trying to get Tailscale to run as a systemd user level service, and ideally Tailscale would be running as root so it can make the necessary networking changes.

This simply installs Tailscale enabling quick and easy setup.

Install command for testing:
```
sudo frzr-deploy diericx/chimeraos:unstable
```

### DIFF
--- a/manifest
+++ b/manifest
@@ -131,6 +131,7 @@ export PACKAGES="\
 	sshfs \
 	steam \
 	sudo \
+	tailscale \
 	tar \
 	tree \
 	ttf-liberation \


alkazar:
Installed size of tailscale is 35 MB which is not too bad, but this seems like a very niche use case. Unless I see more demand for it I would not include it.

diericx:
@alkazar Ok sounds good, thanks for your input! I was thinking about this the past few days and was worried this might be a bloat-y change as it doesn't really provide value to the core os. I'll try to find a better work around and in the meantime I'll close this PR and open an issue to track my progress.

PR #841, opened 2024-01-06
Add cpupower and remove gamemode daemon

ruineka:
The gamemode daemon does not appear to be helping us as much as we may like and adding the cpupower tool can give us more fine controls over how the system manages the power budgets.


pastaq (approved):
Good idea

PR #842, opened 2024-01-06
Add antimicrox to map gamepad to kbm inputs

ruineka:
This utility should prove very useful for mapping any gamepad to control the keyboard and mouse events when using the desktop. It supports profiles that can be created so any gamepad should easily be configurable including external controllers.

More importantly this utility supports Wayland which up to this day the Steam Input solution does not support.

### DIFF
--- a/manifest
+++ b/manifest
@@ -177,6 +177,7 @@ export PACKAGE_OVERRIDES="\
 # Which is often the same as the package name but it can be different.
 # Check on the AUR webpage if you are unsure
 export AUR_PACKAGES="\
+	antimicrox \
 	ayaled \
 	ayn-platform-dkms-git \
 	bcm20702a1-firmware \


ruineka:
> Why not use the flatpak for this?

I think it makes more sense as a system package instead of a flatpak because we'd be shipping it pre installed.

pastaq:
Total Download Size:   12.21 MiB
Total Installed Size:  76.77 MiB

Seems reasonable. Technically I'd say this is a temporary solution. The plan in the future will be to have a "Desktop" mode for HanDPad/HanDBus that is activated automatically when the desktop is started. This is a good holdover method in the meantime.

Does it need to be enabled as a user service or startup program? Can we preconfigure it? How will the users interact with this?

ruineka:
> Total Download Size: 12.21 MiB Total Installed Size: 76.77 MiB
> 
> Seems reasonable. Technically I'd say this is a temporary solution. The plan in the future will be to have a "Desktop" mode for HanDPad/HanDBus that is activated automatically when the desktop is started. This is a good holdover method in the meantime.
> 
> Does it need to be enabled as a user service or startup program? Can we preconfigure it? How will the users interact with this?

You can run it as a startup program with gnome to open automatically and we can provide profiles to be auto loaded. The only issue I'm seeing is I can't seem to get it to run in the background with Gnome.. Not a big deal overall, but it would be a lot better if it did.

PR #845, opened 2024-01-11
Disable amd_pstates by default

ruineka:
The experience with pstates has been very inconsistent and usually results in the CPU being set to maximum frequency even when it's not needed. Some devices have the opposite problem where the CPU will scale down to the lowest possible frequency no matter what. I'm sure some devices may benefit from pstates, but with my testing I don't see that being the majority.

The kernel tries to default to amd-pstate-epp on the 5800U devices and it's much worse than using apcifreq.


pastaq (approved):
As a bonus we'll get CPU Boost toggling again on AMD devices.

PR #855, opened 2024-01-25
Update manifest for rclone

mdeguzis:
https://archlinux.org/packages/extra/x86_64/rclone/

Having rclone built in is super handy for making cloud saves/scripting to backup my game saves and other files.

### DIFF
--- a/manifest
+++ b/manifest
@@ -125,6 +125,7 @@ export PACKAGES="\
 	python \
 	python-inotify-simple \
 	retroarch \
+	rclone \
 	rsync \
 	smbclient \
 	sof-firmware \


pastaq:
How big is the install on disk?

alkazar:
The package is 88 MB installed. It feels a bit too big for the limited use case.

ruineka:
Rclone is extremely useful and is used with emudeck to sync saves to the cloud. I'd argue that this is totally worth it.

ruineka:
> Emudeck doesn't come with a bundled rclone binary?

I'd have to check.

alkazar:
It would have to unless SteamOS includes it?

npaladin2000:
While I agree it's really handy to have it's also easy enough to just drop the binary in your home directory and run it from there.

pastaq:
Before I close this, are we in consensus that we don't want to include it for now?

alkazar:
I think so?

PR #873, opened 2024-02-17
Add AUR Package Build Exit Code Check

chenx-dust:
To prevent failed build task continuing, introduce exit code check.

Example:
<https://github.com/ChimeraOS/chimeraos/actions/runs/7939157980/job/21682894843>
This job failed to package but the whole task still continued.


alkazar:
Nice catch. Thank you!

PR #885, opened 2024-03-07
Replace Ardour with Qtractor

NeroReflex:
This allows sparing more than 100MB of space on the final resulting image while retaining the needed ladspa-host functionality required by chimera quirks package to make some hardware work correctly.

Since I don't own such hardware I couldn't test if the substitution works as intended: someone has to try that before merging this PR.

### DIFF
--- a/manifest
+++ b/manifest
@@ -125,6 +125,7 @@ export PACKAGES="\
 	pulsemixer \
 	python \
 	python-inotify-simple \
+	qtractor \
 	retroarch \
 	rsync \
 	smbclient \


NeroReflex:
Ooohhh I haven't noticed. Guess I will fork fork on my account. No idea why that happened: I used the github GUI button

pastaq:
A little more background, the device-quirks package pulls in `swh-plugins`, which relies on `ladspa-host` as a dependency. Installing with --noconfirm through github actions pulls in the first provider alphabetically, which is `arduor`. Maybe this would be a better fit as a depends/opt-depends in the pkgbuild? 

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=chimeraos-device-quirks-git

NeroReflex:
> Could you clean up the commits? Bit silly change for 3 commits

I have re-done the forking and sent out another PR for merge: https://github.com/ChimeraOS/chimeraos/pull/888

NeroReflex:
> A little more background, the device-quirks package pulls in `swh-plugins`, which relies on `ladspa-host` as a dependency. Installing with --noconfirm through github actions pulls in the first provider alphabetically, which is `arduor`. Maybe this would be a better fit as a depends/opt-depends in the pkgbuild?
> 
> https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=chimeraos-device-quirks-git

If the package really only needs ```swh-plugins``` then listing qtractor there will surely solve the ardour problem but limits how the package can be used. Since that package is available on the AUR I would avoid such move and solve the problem (only relevant to who is concerned with space) at the distro level rather than solving a distro-level problem at package-level. An opt-depend as I see it means "you get additional functionalities if you install this" and this is not the case since ardour should be working exactly as qtractor.

PR #886, opened 2024-03-07
Add xrandr to mangohud dependencies

NeroReflex:
Attempting to build mangohud in a clean chroot with no libxrandr fails as it's missing some needed headers.

Add libxrandr as a dependency.

### DIFF
--- a/pkgs/mangohud/PKGBUILD
+++ b/pkgs/mangohud/PKGBUILD
@@ -7,7 +7,7 @@ arch=('x86_64')
 makedepends=('dbus' 'gcc' 'meson' 'python-mako' 'libx11' 'lib32-libx11' 'git' 'pkgconf' 'vulkan-headers' 'appstream'
              'libxnvctrl'
 )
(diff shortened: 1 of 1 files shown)


NeroReflex:
In file included from ../mangohud/src/app/main.cpp:27:
/usr/include/GLFW/glfw3native.h:119:12: fatal error: X11/extensions/Xrandr.h: No such file or directory
  119 |   #include <X11/extensions/Xrandr.h>
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.

NeroReflex:
Doing some digging brought up this: https://gitlab.archlinux.org/archlinux/packaging/packages/glfw/-/blob/15290026dd3dfb55ac4c580d565af021c2202f3a/PKGBUILD

They marked libxrandr in makedepends, I am unsure if this is a runtime dependency (probably is not) so in this case listing it in makedepends would be a better fit?

PR #888, opened 2024-03-09
Replace Ardour with Qtractor

NeroReflex:
This allows sparing more than 100MB of space on the final resulting image while retaining the needed ladspa-host functionality required by chimera quirks package to make some hardware work correctly.

Since I don't own such hardware I couldn't test if the substitution works as intended: someone has to try that before merging this PR.


pastaq (approved):
Audio phase fix for AYANEO AIR still functions with this change

PR #889, opened 2024-03-09
Btrfs compression

NeroReflex:
Enable btrfs compression at build-time as well as run-time.

The image boots and game just fine and I haven't seen any noticeable slowdown or system load in any of my device.

Chosen compression levels are to achieve minimal image size when building in github and installing to the end-user system and fast write speed for the user (since subvolumes are mounted with a much less aggressive compression ratio).

### DIFF
--- a/build-image.sh
+++ b/build-image.sh
@@ -44,7 +44,7 @@ mkdir -p ${MOUNT_PATH}
 
 fallocate -l ${SIZE} ${BUILD_IMG}
 mkfs.btrfs -f ${BUILD_IMG}
-mount -t btrfs -o loop,nodatacow ${BUILD_IMG} ${MOUNT_PATH}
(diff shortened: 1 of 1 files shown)


alkazar:
My understanding was that compression requires copy on write, which we have disabled.

NeroReflex:
> My understanding was that compression requires copy on write, which we have disabled.

I haven't read that. Can you point me to it? Beside is it disabled in github ci too?

honjow:
A glaring mistake, it should be "compress-force" instead of "force-compress."

pastaq (changes requested):
https://btrfs.readthedocs.io/en/latest/ch-mount-options.html#btrfs-specific-mount-options

NeroReflex:
> A glaring mistake, it should be "compress-force" instead of "force-compress."

Thank you for spotting that!

@pastaq and @alkazar I have also removed nodatacow to effectively allow compression and defragmented the filesystem.

A subvolume created without nodatacow can be mounted with that flag and vice-versa.

PR #914, opened 2024-03-30
mkinitcpio: use systemd in initramfs

NeroReflex:
Add systemd in initramfs: this changes the old init program with systemd and also provides a default keymap and font for default initramfs images.

### DIFF
--- a/rootfs/etc/mkinitcpio.conf
+++ b/rootfs/etc/mkinitcpio.conf
@@ -3,4 +3,4 @@
 MODULES=(dm_mod ext4 sha256 sha512 overlay)
 BINARIES=()
 FILES=()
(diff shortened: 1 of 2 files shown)


alkazar:
What problem does this solve?

NeroReflex:
This is part of frzr rework: is required to not specify rootflags in bootloader thus making it much easier to switch kernels and remove a failure point in update while increasing the atomicity of the update procedure itself.

alkazar:
Awesome, thanks. Can it be merged safely now?

NeroReflex:
> Awesome, thanks. Can it be merged safely now?

I don't really know. I have been testing it in my ally and main rig and it works well, however I am unsure frzr-etc will be compatibile. Reworking fstab to include the overlayfs with x-initrd will be much safer option in this case.

PR #929, opened 2024-04-07
Add ntfs-3g to packages

GasparAM:
Most of the users who use some kind of external hard drive have them formatted to ntfs. Adding the following package will allow users to use external hard drives for storing games, media and allowing transfer of files from windows PC's.


NeroReflex:
The current ntfs driver in the kernel should be able to mount drives in r/w, what is the problem exactly?

PR #943, opened 2024-04-30
feat: Introduce workaround for Steam Input in wayland session.

### LINKED ISSUE #700: ROG Ally (Steam Controller) invisible mouse, and fix
Hey, not sure if this is the right place for this, but I found an interesting (and apparently very old) bug.

1. On the ROG Ally, when you attempt to use Steam Input in Desktop mode, the mouse cursor is invisible. 
2. In order to get a visible cursor, I had to attach another mouse. 
3. Once the cursor is visible, if you go back to using the joystick, the selection point moves, but the cursor itself does not. 

It looks like it has been a long-time issue for the Steam Controller on Wayland. 

https://github.com/ValveSoftware/steam-for-linux/issues/4738
https://github.com/ValveSoftware/steam-for-linux/issues/6615

I was able to get it working with the following project: 
https://github.com/Supreeeme/extest

chenx-dust:
Fix #700

### DIFF
--- a/manifest
+++ b/manifest
@@ -323,7 +323,7 @@ postinstallhook() {
 		xargs -I {} sh -c "echo NoDisplay=true >> {}"
 
 	# force -steamdeck option in desktop mode to prevent constant steam updates
-	sed -i 's,Exec=/usr/bin/steam-runtime,Exec=/usr/bin/steam-runtime -steamdeck,' /usr/share/applications/steam.desktop
+	sed -i 's,Exec=/usr/bin/steam-runtime,Exec=/usr/bin/steam-desktop-mode,' /usr/share/applications/steam.desktop
 
 	# set permissions for intel_gpu_top and mangohud
 	setcap cap_perfmon=+ep /usr/bin/intel_gpu_top
--- a/manifest
+++ b/manifest
@@ -211,6 +211,7 @@ export AUR_PACKAGES="\
 	legendary \
 	lib32-gamescope-plus \
 	lib32-gperftools \
+	lib32-extest \
 	libretro-dosbox-pure-git \
 	libretro-opera-git \
 	libretro-prosystem-git \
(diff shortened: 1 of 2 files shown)


chenx-dust:
Seem that it still has problem for autostarted steam. I should load `libextest` whether it's a wayland session or not. Is that a good idea?

rootfs/usr/bin/steam-desktop-mode
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/bash
```
alkazar, line 1: This is unnecessary. We already force Steam to launch in Deck mode.

manifest
```diff
@@ -321,8 +322,9 @@ postinstallhook() {
 		grep -v steam.desktop |
 		xargs -I {} sh -c "echo NoDisplay=true >> {}"
 
-	# force -steamdeck option in desktop mode to prevent constant steam updates
-	sed -i 's,Exec=/usr/bin/steam-runtime,Exec=/usr/bin/steam-runtime -steamdeck,' /usr/share/applications/steam.desktop
```
alkazar, line 325: Why remove this?

manifest
```diff
@@ -321,8 +322,9 @@ postinstallhook() {
...
+	# use script to fix steam input in wayland mode
+	sed -i 's,Exec=/usr/bin/steam-runtime,Exec=/usr/bin/steam-desktop-mode,' /usr/share/applications/steam.desktop
+	sed -i 's,Exec=/usr/bin/steam,Exec=/usr/bin/steam-desktop-mode,' /etc/xdg/autostart/steam.desktop
```
alkazar, line 327: I don't want to start Steam automatically. That seems quite heavy handed.

alkazar:
Hi, sorry for a lack of communication. I forgot to follow up. We plan to have a different solution that does not rely on Steam for navigating the desktop with a gamepad.

PR #958, opened 2024-05-18
Migrate to the new frzr

NeroReflex:
This PR is meant to have the system compatible with the new frzr, while also retaining the ability to be installed with an older version of frzr.

Amongs its goals there is the ability to apply old-frzr quirks (like renaming initramfs-linux-chimeraos.img into initramfs-linux.img only whene such a change is actually required)


NeroReflex:
This PR is part of another merged one.

PR #965, opened 2024-06-03
Use archlinux-aur as Docker base image and install InputPlumber as a dependency to build OpenGamepadUI

ShadowApex:
This change uses the [archlinux-aur](https://github.com/greyltc-org/docker-archlinux-aur) image as the base for building in Docker. It allows us to install packages from the AUR in the container that may be required to build other AUR packages (like OpenGamepadUI)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,51 +1,54 @@
-FROM archlinux:base-devel
+FROM greyltc/archlinux-aur:latest
 LABEL contributor="shadowapex@gmail.com"
 COPY rootfs/etc/pacman.conf /etc/pacman.conf
 RUN echo -e "keyserver-options auto-key-retrieve" >> /etc/pacman.d/gnupg/gpg.conf && \
-    # Cannot check space in chroot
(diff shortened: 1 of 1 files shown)


pastaq (approved):

Dockerfile
```diff
@@ -1,51 +1,54 @@
...
+  python-setuptools \
+  python-wheel \
+  sudo \
+  && \
+  pacman --noconfirm -S --needed git && \
+  echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
+  useradd build -G wheel -m && \
+  su - build -c "git clone https://aur.archlinux.org/pikaur.git /tmp/pikaur" && \
+  su - build -c "cd /tmp/pikaur && makepkg -f" && \
+  pacman --noconfirm -U /tmp/pikaur/pikaur-*.pkg.tar.zst
```
ShadowApex, line 30: These changes are just from a formatter

PR #975, opened 2024-06-19
Use dracut instead of mkinitcpio

NeroReflex:
The following considerations and observations have been made:

- dracut is more configurable
- dracut is the only initramfs generator that will work with the "frzr kernel" utility
- a dracut-generated initramfs is 38MB, while mkinitcpio is about 120MB
- the included dracut configuration avoids building the fallback, thus sparing build time


NeroReflex:
The current frzr is tightly coupled to device-quirks and mkinitcpio: in fact frzr lists mkinitcpio among its dependencies.

This PR is to be merged as part as the frzr refactor.

pastaq (approved):
Looks good. @NeroReflex can you list all the other related patches in this issue? Having a central location to track all the moving pieces will help.

PR #977, opened 2024-06-20
fix(InputPlumber): Set intercept mode of InputPlumber to none in desktop

pastaq:


pastaq:
This might not be the best solution. We might need a polkit to restart the service instead to clear various things like the target devices, gamepad profile, etc.

PR #987, opened 2024-07-25
wol: move the udev rule on its own package

NeroReflex:
- make more explicit ethtool is pulled in for a specific udev rule
- make easier for a user to disable wake-on-lan
- bind ethtool to the specific functionality requireing it


NeroReflex:
This is now handled differently: see https://github.com/ChimeraOS/chimeraos/pull/1016

PR #988, opened 2024-07-25
ntfs: move the udev rule on its own package

NeroReflex:
Make it easier for an user to spot chimeraos-specific customizations by moving files in theirs own package.


pastaq:
Most of our udev rules are already moved to device quirks unless they are package specific

NeroReflex:
> Why not merge all this stuff into one package? Similar to steamos-customizations?

To keep them separate has its value as it's easy to understand what package does what, and someone looking at the repo can easily find out where configurations are coming from.

Plus, may the need arise to remove any of that it will be easier this way.

NeroReflex:
This is now handled differently: see https://github.com/ChimeraOS/chimeraos/pull/1016

PR #989, opened 2024-07-25
Attempt to add Nvidia support back

ruineka:
The explicit sync support has hit stable with the Nvidia drivers and it is now possible for us to begin to support Nvidia again.

### DIFF
--- a/manifest
+++ b/manifest
@@ -121,6 +121,12 @@ export PACKAGES="\
 	nfs-utils \
 	noto-fonts-emoji \
 	nss-mdns \
+	nvidia \
+	opencl-nvidia \
+	lib32-opencl-nvidia \
+	nvidia-utils \
+	lib32-nvidia-utils \
+	nvidia-prime \
 	openal \
 	openrazer-daemon \
 	openssh \


pastaq:
Do the icd's not longer cause long timeout on AMD hardware when starting up?

ruineka:
> Do the icd's no longer cause long timeout on AMD hardware when starting up?

It might still do that, I will need to do testing to see how things stand now. It's been a long while.

alkazar:
Should we use NVK instead?

ruineka:
> Should we use NVK instead?

It's limited to the 20 series and newer and it isn't really ready for gaming in general let alone trying to use it with game mode. Nvidia with explicit sync should no longer have the graphical issues we've had to deal with in the past.

Samsagax:
Make sure you also revert this on gamescope. That comes with all different can of worms: https://github.com/ChimeraOS/gamescope/commit/7865b34c5cd61fa5cc5428ace614e4551fabb6ec

matte-schwartz:
You actually don’t need explicit sync enabled in Gamescope for nvidia to work. It's still not fixed on AMD cards so I wouldn't recommend it.

ruineka:
> You actually don’t need explicit sync enabled in Gamescope for nvidia to work. It's still not fixed on AMD cards so I wouldn't recommend it. 

Work without the graphical corruption that was present in the past? It would "work" in the past, but the experience was bad overall.

matte-schwartz:
I have not noticed a difference in my testing, but I can only attest for 4xxx series cards.

ruineka:
> Make sure you also revert this on gamescope. That comes with all different can of worms: https://github.com/ChimeraOS/gamescope/commit/7865b34c5cd61fa5cc5428ace614e4551fabb6ec

I don't need to revert this, I just need to make it toggle-able.

ruineka:
> I have not noticed a difference in my testing, but I can only attest for 4xxx series cards.

Thanks for the information.

PR #991, opened 2024-07-25
Fix installation in non-nvme and non-mmcblk device types

NeroReflex:
The installation was failing from devices like /dev/sda but was working on /dev/nvme and /dev/mmcblk disks due to a wrong regex used to extract the disk from the partition.

### DIFF
--- a/rootfs/usr/lib/frzr.d/install-0003-root_part_gpt.migration
+++ b/rootfs/usr/lib/frzr.d/install-0003-root_part_gpt.migration
@@ -23,16 +23,17 @@ post_install() {
     echo "Fetching partition number of partition '$subcmd'"
     local part_number=$(echo "${subcmd: -1}")
 
-    local disk=$(echo "$subcmd" | sed 's/p[0-9]*$//')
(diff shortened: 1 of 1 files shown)


alkazar:
This is a migration script... it shouldn't be triggered by a new install and we should probably not edit it. There must be something going wrong elsewhere.

NeroReflex:
> This is a migration script... it shouldn't be triggered by a new install and we should probably not edit it. There must be something going wrong elsewhere.

This migration script is what flags the main btrfs partition as "Linux rootfs x64_86" instead of the generic "Linux Filesystem" and I intended it to run because it's strictly necessary when coming from a pre-refactor frzr.

I confirmed (just now, on my branch)  that this commit is what solves the (new, due to frzr refactor) installation problem on devices like /dev/sda2

alkazar:
Oh, I see. It is a new migration script then... does this mean we need another release that has this updated script and that does not have the refactor in order for the migration to happen correctly?

NeroReflex:
> Oh, I see. It is a new migration script then... does this mean we need another release that has this updated script and that does not have the refactor in order for the migration to happen correctly?

No, the refactor, with the new set of migrations is fully able to move back and forth with no restrictions :)

PR #992, opened 2024-07-25
Image compression

NeroReflex:
To spare the most possible space for the github release use the highest possible compression on the final image.


alkazar:
The original value was taken as a balance between size and decompression time. How much longer does decompression take?

NeroReflex:
> The original value was taken as a balance between size and decompression time. How much longer does decompression take?

I honestly have no idea because it's impossible to know without measuring it. Plus this is going to be used with the refactored frzr:
The refactored frzr can be much faster than the old one because it does not need to rebuild initramfs or can be slower because it can also perform a disk check... and I don't know what default is better.

pastaq:
Considering how close we are to the 2GB limit, especially if we include Nvidia, I think this is a good idea. Decompression happens in the background anyway for updates, which is the bulk of the time spend decompressing in the long term. A slightly longer install is a good compromise IMO.

PR #1000, opened 2024-07-28
Build system updates

BoukeHaarsma23:
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


NeroReflex:
I am concerned with the don't compreda artifacts.
Would kindly ask for some evidence it does good.
Right now we use xz -8 and it can be compressed more for sure (see my other PR) so data is surely not-non-compressible right now

PR #1016, opened 2024-08-25
Move udev rules

NeroReflex:
This PR moves udev rules and additional dependencies for those udev rules into device-quirks package. 

Also accomplish what is requested here: https://github.com/ChimeraOS/device-quirks/pull/114

### DIFF
--- a/manifest
+++ b/manifest
@@ -31,7 +31,6 @@ export PACKAGES="\
 	dosbox \
 	efibootmgr \
 	epiphany \
-	ethtool \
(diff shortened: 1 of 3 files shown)


manifest
```diff
@@ -31,7 +31,6 @@ export PACKAGES="\
 	dosbox \
 	efibootmgr \
 	epiphany \
-	ethtool \
```
pastaq, line 34: This change is related how?

PR #1021, opened 2024-09-22
Spatial Audio Support via Passthrough

pelasgus:
- Rough implementation
- Have not accounted for CI


pastaq:
This is an interesting change that looks to add good value. That k you for spending the time on resolving this. I think it would make more sense to be added as part of the device quirks package rather than as a build time globally enabled quirk. The pastaq/new_frzr branch would be the best target for it as that is where we've refactored a lot of quirks and it has the ability to add and remove these kinds of things.

I have a few considerations for you to consider as well when moving it over:
- Is there a way to detect when this is necessary during installation and upgrades? It would be preferable to only apply such things to affected hardware. If not, it needs to be manually triggerable.
- Will this conflict with the pipewire or wireplumber configurations we use for some handhelds, and if it does how to we mitigate those issues?
- It needs to be capable of determining if the quirk needs to be updated if it already exists, and it needs to be removable with a single command.

pelasgus:
- Ported to wireplumber
- Conditional logic enforced
- Monitors nodes; if node with hdmi then passes through
- This is the expected behaviour really, if you have a device connected on hdmi you want audio to be passed through, not pre-processed; a la consoles

Alas, if you want to enforce further conditional logic, then I ll rename directories from /etc/wireplumber/scripts to /usr/share/wireplumber/scripts and move it to the device quirks repo. This should not be a device quirk though, its expected default experience from a console.

pastaq:
> Alas, if you want to enforce further conditional logic, then I ll rename directories from /etc/wireplumber/scripts to /usr/share/wireplumber/scripts and move it to the device quirks repo. This should not be a device quirk though, its expected default experience from a console.

The purpose of the device quirks package is precisely to add these kinds of changes so that the default console experience is good, but only applying them when needed. I think it would fit better there. Also, we avoid installing anything to /etc unless absolutely necessary as that is preserved during updates which means we could override user intent or it could still be present if we delete something later.

Issue #556, opened 2023-04-02
Xbox controller (s) Bluetooth won't connect until navigate to Bluetooth page


eyecreate:
Xbox controller won't find pc until navigating to steam Bluetooth settings. Expected result was being able to reconnect controller after turning on/reboot without another input device to navigate to Bluetooth settings.

edwin-jones:
I had the same issue on chimeraOS 39 - I also couldn’t bring up the gamescope options with Xbox logo + A unless plugged in via usb. I haven’t been able to test 41 due to #516

pastaq:
> I had the same issue on chimeraOS 39 - I also couldn’t bring up the gamescope options with Xbox logo + A unless plugged in via usb. I haven’t been able to test 41 due to #516

this sounds like three issues

edwin-jones:
Fair point, edited to keep focus.

maddox:
I'm having this issue as well on a second build. My first build does not have this issue. 

FWIW, the build that works has a Bluetooth 5.1 card on it, and the one that doesn't work well has 4.2.

pparidans:
Same issue here with an Xbox Series controller.

The controller is fully updated (just in case). And the current Wi-Fi/Bluetooth card is an Intel AC 7265 (Bluetooth 4.2).

Will try with another card (Intel AX 200) supporting Bluetooth 5.2 and report the result back here...

NeroReflex:
Does adding 
'''
KernelExperimental=true
Experimental=true
'''
in section [General] of file /etc/bluetooth/main.conf change something?

pparidans:
Some feedback :
- I did upgrade the Wi-Fi/Bluetooth card to an Intel AX 200 and no change (actually the Bluetooth is even less reliable :grimacing: )
- Updated to Chimera 43, but no change either...

My current "solution" is... to use the Microsoft XBOX Wireless Adapter :shrug:

Samsagax:
Original Xbox controllers work best with their original dongle. I'm yet to see any of them work on regular bluetooth.

ChaosBlades:
My Xbox One S bluetooth controller was not seen by DeckUI but was by Gnome DE. Worked without issue on HoloISO.

Daasin:
Similar issue here and it does nothing once it does connect, I go to controllers and it says no controller detected for an Xbox Elite controller

pastaq:
[...]

PR #1029, opened 2024-10-09
Bluetooth auto reconnect service

RowlaxX:
# Motivation

Hi, after installing ChimeraOS, I was unable to connect my Xbox Series controller until I navigated to the bluetooth settings. It was very similar to [this issue](https://github.com/ChimeraOS/chimeraos/issues/556)

# Investigation

After searching the cause of this problem, it seems that a Bluetooth scan was triggered while navigating to the Gnome Bluetooth Settings, and disabled when leaving that page.  

Later, I was able to reconnect my controller by using the command `bluetoothctl -t -1 scan on` without navigating to the bluetooth settings

# Solution

This PR add a simple systemd service that periodically scan for previously connected Bluetooth devices

I am now able to automatically reconnect my controller in the following cases : 
- On startup even before the session is loaded
- When I unplug & replug my bluetooth dongle
- When my controller is disconnected due to idle or battery and I press the guide button

# Mention

It seems that Bluez allow for periodic scans in the config via `ScanIntervalAutoConnect` & `ScanWindowAutoConnect` parameters ([link](https://github.com/Vudentz/BlueZ/blob/master/src/main.conf)).

This could be a way cleaner solution, but I was unable to make them work.

# Hardware & Software

- Xbox Series Official Wireless Controller
- my bluetooth dongle (https://www.amazon.fr/dp/B09TT7SXHY?ref=ppx_yo2ov_dt_b_fed_asin_title)
- NVidia Geforce 1080TI
- ChimeraOS 47 (b88d2bf) [UNSTABLE]

# Note

This is my first PR for Chimera so I may not have respected some design & architectural choice

### DIFF
--- a/manifest
+++ b/manifest
@@ -249,6 +249,7 @@ export SERVICES="\
 	NetworkManager \
 	avahi-daemon \
 	bluetooth \
+	bluetooth-auto-connect \
 	bluetooth-workaround \
 	fstrim.timer \
 	haveged \
--- a/rootfs/etc/systemd/system/bluetooth-auto-connect.service
+++ b/rootfs/etc/systemd/system/bluetooth-auto-connect.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=A service that periodicaly scan & connect for previously connected bluetooth devices
+After=bluetooth.service
+
+[Service]
+ExecStart=/usr/bin/bluetoothctl -t 3 scan on
+Restart=Always
+RestartSec=1
+
+[Install]
+WantedBy=multi-user.target


alkazar:
Thanks for the work, I am a bit confused as to why this is necessary.

If the controller is paired, it should be able to autoconnect without this service.

RowlaxX:
It appears that Bluez don't support autoconnect for some bluetooth controller unless there is a bluetooth scan occuring. It was the case with the bluetooth dongle mentioned above.

However, I tried to run a fresh chimeraos with another bluetooth dongle ([this one](https://www.amazon.fr/dp/B098K3H92Z?ref=ppx_yo2ov_dt_b_fed_asin_title)) and didn't got any issue

This service is a fix for controllers that need a scan for connecting devices

alkazar:
I see, so it is hardware dependent. Thank you.

I am not sure if it is a good idea to be scanning constantly like that. I will have to do some research and think about it.

kharenis:
@alkazar Have you had a chance to consider this further? I've had to implement it myself locally to resolve the same issue as there seems to be no other viable solution for me.

I implemented the following;
`[Unit]
Description=A service that periodicaly scan & connect for previously connected bluetooth devices
After=bluetooth.service

[Service]
ExecStart=/usr/bin/bluetoothctl -t 3 scan on
Restart=always
RestartSec=15`

RowlaxX:
I bought a pci bluetooth/wifi card and this problem has been resolved for me. It is hardware dependent as far as I can tell.

Maybe a nice compromise could be to add this service for the next release and keeping it disabled by default. We can also add an option to the web UI to enable/disable this service in case some user have this problem.

I can start working on that if needed @alkazar

jossmusix:
@RowlaxX can you share which card you bought? Im having bt troubles also (with a fenvi card)

RowlaxX:
This one :
[TP-Link WiFi 6 PCIe WiFi Card AX1800 Archer TX20E](https://amzn.eu/d/6Yavhr3)

jossmusix:
@RowlaxX Bonne soirée et passe un excellent dimanche soir!

alkazar:
As a compromise, I was thinking what if we run this service for the first few minutes after startup?

kharenis:
That sounds like a good compromise to me. Sadly a PCIe card is out of the question for me as I don't have enough slots.

arthurlutz:
I think having the service integrated, disabled by default and an option (with a warning about constant scanning)  would be a nice work around so that other users don't have to go into the depths of issues/wiki/discord to find a fix for this and supporting existing hardware instead having to buy new hardware is always a good thing. 

Would there be a nice privacy-preserving way to report back the bluetooth hardware used so that in the future this could be automatically enabled when certain hardware setups are detected ?

